PR #1, opened 2018-12-13
Moving hostpath drvier from kubernetes-csi/drivers monorepo

vladimirvivien:
This PR is to setup and migrate the Hostpath CSI driver from the github.com/kubernetes-csi/drivers mono repo to its own repository.


pohly:
I think it would be good to use the same directory layout as in other stand-alone components (like external-provisioner) with `pkg` for packages and `cmd/hostpathplugin/main.go` for the binary.

I'd also like to see the history of the drivers imported. That way it'll be easier to review how this version of the plugin differs from the original one and (in the future) do code archeology if that ever becomes necessary.

I can push an updated branch with those changes if you want.

pohly:
As it's so easy, I went ahead and prepared a commit which is identical to this PR, but with full history:
https://github.com/pohly/csi-driver-host-path/tree/transfer-hostpath

In that branch, `git annotate pkg/hostpath/hostpath.go` and `git log` shows the original authors and history.

The way how I did that is:
```
git remote add drivers git@github.com:kubernetes-csi/drivers
git fetch drivers
GIT_AUTHOR_EMAIL=vivienv@vmware.com GIT_AUTHOR_NAME="Vladimir Vivien" git commit-tree -p 003de657d5e0a25583b9e90b69bf193926700c11 -p drivers/master -m "Adds source code"  58c663832ac2587e222495e46f6513d8bd78060c
```

The two hashes come from "git show --format=raw" for this PR.

@vladimirvivien if you like this, feel free to fetch my branch and force-push it here.

vladimirvivien:
@pohly super thankful for the great review.  

Let me get it all done and then I will open PRs to do your suggestions 1) bring over history and 2) match dir structure under cmd/.  

That way I am not fighting unnecessary rebase conflicts.

pohly:
@vladimirvivien we need to bring in history before merging this PR. But feel free to finish your work and then I can splice in history once more before we merge.

Gopkg.toml
```diff
@@ -0,0 +1,78 @@
+
+# Gopkg.toml example
+#
+# Refer to https://github.com/golang/dep/blob/master/docs/Gopkg.toml.md
+# for detailed Gopkg.toml documentation.
+#
+# required = ["github.com/user/thing/cmd/thing"]
+# ignored = ["github.com/user/project/pkgX", "bitbucket.org/user/project/pkgA/pkgY"]
+#
```
pohly: Please prune the `vendor` dir. See https://github.com/kubernetes-csi/drivers/pull/127

pohly: The pruning of the vendor dir should get squashed into your "Adds vendor dir" commit, because that way we never end up with the full set of files in the master branch.

vladimirvivien:
@pohly I pruned the vendor.  All yours to add the history.

pohly:
```
$ dep ensure
Warning: the following project(s) have [[constraint]] stanzas in Gopkg.toml:

  ✗  github.com/spf13/cobra
  ✗  github.com/stretchr/testify
```

I'm going to fix this by removing the constraints.

@vladimirvivien please replace your branch content with https://github.com/pohly/csi-driver-host-path/tree/b5840132190aacc764cd016137597eb16dead852

It's content is identical to what's here now, except for the Gopk change:
```
$ git diff b5840132190aacc764cd016137597eb16dead852..44f210d818ef40649e0c98707e0dabd268850c88 | diffstat
 Gopkg.lock |   52 ++++++++++++++++++++++++++++++++++++++++++++++------
 Gopkg.toml |    8 --------
 2 files changed, 46 insertions(+), 14 deletions(-)
```

I took the liberty to split out your deployment changes from the "Adds source code" commit. That way it can be reviewed and potentially updated more easily. If you want to amend a commit, you can do it with `git rebase -i 63df35fc1394508f33aa464b236dcbb958e40b7c` without dropping the fake merge.

deploy/deploy-hostpath.sh
```diff
@@ -0,0 +1,25 @@
...
+# This script captures the steps required to successfully
+# deploy the hostpath plugin driver.  This should be considered
+# authoritative and all updates for this process should be
+# done here and referenced elsewhere.
+
+# The script assumes that kubectl is available on the OS path 
+# where it is executed.
+
+K8S_RELEASE=${K8S_RELEASE:-"release-1.13"}
+CSI_RELEASE=${CSI_RELEASE:-"release-1.0"}
```
pohly: I think these two variables are a bit misleading. Instead of selecting one particular, well-tested release they reference the latest code in certain release branches.

I understand that this is easier to implement and review, but is it really what we want?

I personally would prefer a solution where for each of the sidecars we install exactly the RBAC rules released together with that sidecar image.

There's also no guarantee that we will always have the same release branch for all apps. For example, we might release an external-provisioner 1.1.0 on a release-1.1 branch while the rest need no major new release and thus stay on the release-1.0 branch.

deploy/hostpath/csi-hostpath-rbac.yaml
```diff
@@ -0,0 +1,46 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: csi-node-sa
```
pohly: I think this entire file is obsolete.

The hostpath driver should use the RBAC rules required by the csi-node-driver-registrar. At least on master, that file was removed (https://github.com/kubernetes-csi/node-driver-registrar/issues/7#issuecomment-449414820), but the release-1.0 still has it and therefore it should be used.

pohly: https://github.com/kubernetes-csi/node-driver-registrar/blob/release-1.0/deploy/kubernetes/rbac.yaml

app/main.go
```diff
@@ -0,0 +1,46 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
```
pohly: Please move to a `cmd/hostpathplugin` directory.

deploy/deploy-hostpath.sh
```diff
@@ -0,0 +1,25 @@
...
+BASE_DIR=$(dirname "$0")
+
+# apply CSIDriver and CSINodeInfo API objects
+kubectl apply -f https://raw.githubusercontent.com/kubernetes/csi-api/${K8S_RELEASE}/pkg/crd/manifests/csidriver.yaml --validate=false
+kubectl apply -f https://raw.githubusercontent.com/kubernetes/csi-api/${K8S_RELEASE}/pkg/crd/manifests/csinodeinfo.yaml --validate=false
```
pohly: This is not always needed, is it?

Worse, can it fail when the CRDs were already installed, which (if I'm not mistaken) is how clusters now should be brought up?

deploy/deploy-hostpath.sh
vladimirvivien: @pohly I am taking a simple approach, in the next push, to just create a variable for each component.  That way user of the script can set versions independently.

deploy/deploy-hostpath.sh
vladimirvivien: @pohly I don't have a good fix for this. I will add a flag (or try to detect if they are there) to install only when needed.  Thanks for pointing this out.

pohly:
/lgtm

@saad-ali you might have to push this PR directly to master because of the CLA check for some of the commits that this is pulling in from the old repo.

@dims ran into a similar issue with moving code into kubernetes/utils, see https://github.com/kubernetes/utils/pull/55#issuecomment-451591270 for the suggested solution (same as above).

PR #5, opened 2019-01-25
hostpath: Implement ValidateVolumeCapabilities

darkowlzz:
This implements a basic version of ValidateVolumeCapabilities for
hostpath driver and removes this from the skip tests list in the e2e
tests.

Running the e2e tests locally always failed because it was skipped for travisCI environment only.

### DIFF
--- a/pkg/hostpath/controllerserver.go
+++ b/pkg/hostpath/controllerserver.go
@@ -163,7 +163,35 @@ func (cs *controllerServer) ControllerGetCapabilities(ctx context.Context, req *
 }
 
 func (cs *controllerServer) ValidateVolumeCapabilities(ctx context.Context, req *csi.ValidateVolumeCapabilitiesRequest) (*csi.ValidateVolumeCapabilitiesResponse, error) {
-	return nil, status.Error(codes.Unimplemented, "")
+
+	// Check arguments
+	if len(req.GetVolumeId()) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "Volume ID cannot be empty")
+	}
+	if len(req.VolumeCapabilities) == 0 {
+		return nil, status.Error(codes.InvalidArgument, req.VolumeId)
+	}
+
+	if _, err := getVolumeByID(req.GetVolumeId()); err != nil {
+		return nil, status.Error(codes.NotFound, req.GetVolumeId())
+	}
+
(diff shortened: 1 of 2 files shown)


pkg/hostpath/controllerserver.go
```diff
@@ -147,7 +147,35 @@ func (cs *controllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVol
...
+		return nil, status.Error(codes.NotFound, req.GetVolumeId())
+	}
+
+	for _, cap := range req.GetVolumeCapabilities() {
+		if cap.GetMount() == nil && cap.GetBlock() == nil {
+			return nil, status.Error(codes.InvalidArgument, "cannot have both mount and block access type be undefined")
+		}
+
+		// A real driver would check the capabilities of the given volume with
+		// the set of requested capabilities.
```
pohly: I'm wondering whether any "real" driver properly implements this? I was trying to do it recently in my own CSI driver when adapting it to CSI 1.0 (sadly, still not public). It was far from trivial, so it might be good to have a correct example.

gcepd doesn't implement it:
https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/162

This is just an observation, it shouldn't hold back merging this PR.

darkowlzz: My driver does something like this:
```go
	mode := cap.GetAccessMode().GetMode()
	switch mode {
	case csi.VolumeCapability_AccessMode_SINGLE_NODE_WRITER:
		// Supported capability. Continue checking other capabilities.
	case csi.VolumeCapability_AccessMode_SINGLE_NODE_READER_ONLY:
		message = "Volume is not singlenode read only volume"
		break
	case csi.VolumeCapability_AccessMode_MULTI_NODE_MULTI_WRITER,
		csi.VolumeCapability_AccessMode_MULTI_NODE_READER_ONLY,
		csi.VolumeCapability_AccessMode_MULTI_NODE_SINGLE_WRITER:
		message = "Volume is not multinode volume"
		break
	default:
		return nil, status.Errorf(codes.InvalidArgument, "AccessMode %s is not allowed", mode)
	}
```
I'm not sure if that's how it should be done.
Maybe we can have a commented example in this driver if there's no other sample drivers that has different capabilities.

darkowlzz:
Rebased.

PR #7, opened 2019-02-06
add approvers for hostpath csi driver

msau42:


OWNERS
```diff
@@ -3,6 +3,8 @@
 approvers:
 - childsb
 - saad-ali
-reviews:
+- vladimirvivien
+- msau42
+reviewers:
 - childsb
 - saad-ali
```
pohly: Does it make sense to keep @childsb and @saad-ali as reviewers? In other words, are both of you really going to look at PRs against this repo? Same question is relevant also for some of the other kubernetes-csi repos.

PR #9, opened 2019-02-12
Removed csi-comman package dependency

avalluri:
csi-common package is part of kuberntes-csi/drivers which is deprecated and
unmaintained. The main parts used from csi-common are default
{node,identity,controller}server implementations and NonBlockingGRPCServer.

Default server implementations can be integrated to hostpath driver code, this
makes the driver code simple and clean. Lifted over the NonBlockingGRPCServer
code and simplified as per the driver needs.

Signed-off-by: Amarnath Valluri <amarnath.valluri@intel.com>


pkg/hostpath/server.go
```diff
@@ -72,14 +63,14 @@ func (s *nonBlockingGRPCServer) ForceStop() {
...
-	proto, addr, err := ParseEndpoint(endpoint)
+	proto, addr, err := parseEndpoint(endpoint)
 	if err != nil {
 		glog.Fatal(err.Error())
 	}
 
 	if proto == "unix" {
 		addr = "/" + addr
-		if err := os.Remove(addr); err != nil && !os.IsNotExist(err) {
+		if err = os.Remove(addr); err != nil && !os.IsNotExist(err) {
```
pohly: It's cleaner to use `err :=` here, then the if check is independent from the preceding code.

pkg/hostpath/server.go
```diff
@@ -110,3 +101,25 @@ func (s *nonBlockingGRPCServer) serve(endpoint string, ids csi.IdentityServer, c
...
+		if s[1] != "" {
+			return s[0], s[1], nil
+		}
+	}
+	return "", "", fmt.Errorf("Invalid endpoint: %v", ep)
+}
+
+func logGRPC(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+	glog.V(3).Infof("GRPC call: %s", info.FullMethod)
+	glog.V(5).Infof("GRPC request: %+v", req)
```
msau42: These should use protosanitizer.

pkg/hostpath/server.go
```diff
@@ -14,32 +14,23 @@ See the License for the specific language governing permissions and
...
-	// Waits for the service to stop
-	Wait()
-	// Stops the service gracefully
-	Stop()
-	// Stops the service forcefully
-	ForceStop()
-}
-
-func NewNonBlockingGRPCServer() NonBlockingGRPCServer {
+func NewNonBlockingGRPCServer() *nonBlockingGRPCServer {
```
msau42: @pohly do you think this rpc server boilerplate may be useful as a library?

pkg/hostpath/server.go
avalluri: It was trying to fix warning raised by vetshadow, Now i reverted the change, and added //nolint.

pkg/hostpath/server.go
avalluri: done

pkg/hostpath/server.go
pohly: I was thinking about that but then decided against it, because the other sidecar apps are no gRPC servers and thus do not need this code. Other CSI drivers might use it, but I think the function is not big enough to justify sharing it.

pkg/hostpath/server.go
```diff
@@ -110,3 +102,25 @@ func (s *nonBlockingGRPCServer) serve(endpoint string, ids csi.IdentityServer, c
...
+}
+
+func logGRPC(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+	glog.V(3).Infof("GRPC call: %s", info.FullMethod)
+	glog.V(5).Infof("GRPC request: %+v", protosanitizer.StripSecrets(req))
+	resp, err := handler(ctx, req)
+	if err != nil {
+		glog.Errorf("GRPC error: %v", err)
+	} else {
+		glog.V(5).Infof("GRPC response: %+v", resp)
```
msau42: response also should use protosanitizer

msau42:
/approve
Just one comment about protosanitizer
will let @pohly lgtm

pkg/hostpath/server.go
avalluri: @msau42  fixed it now.

avalluri:
@pohly do you think its good enough to merge

PR #13, opened 2019-02-22
Indicate --kubelet-registration-path through env

blaggacao:
When people operate kubelet with `--root-dir=/foo/bar`, those values
need to be changed. In order to (subtly) indicate externalities of --kubelet-registration-path
feed it through an env variable.

### DIFF
--- a/deploy/hostpath/csi-hostpath-plugin.yaml
+++ b/deploy/hostpath/csi-hostpath-plugin.yaml
@@ -31,7 +31,7 @@ spec:
                   apiVersion: v1
                   fieldPath: spec.nodeName
             - name: DRIVER_REG_SOCK_DIR
-              value: /opt/rke/var/lib/kubelet/plugins/csi-hostpath
+              value: /var/lib/kubelet/plugins/csi-hostpath
           volumeMounts:
           - mountPath: /csi
             name: socket-dir
--- a/deploy/hostpath/csi-hostpath-plugin.yaml
+++ b/deploy/hostpath/csi-hostpath-plugin.yaml
@@ -23,13 +23,15 @@ spec:
           args:
             - --v=5
             - --csi-address=/csi/csi.sock
-            - --kubelet-registration-path=/var/lib/kubelet/plugins/csi-hostpath/csi.sock
+            - --kubelet-registration-path=$(DRIVER_REG_SOCK_DIR)/csi.sock
           env:
             - name: KUBE_NODE_NAME
               valueFrom:
                 fieldRef:
                   apiVersion: v1
                   fieldPath: spec.nodeName
+            - name: DRIVER_REG_SOCK_DIR
+              value: /opt/rke/var/lib/kubelet/plugins/csi-hostpath
           volumeMounts:
           - mountPath: /csi
             name: socket-dir


blaggacao:
I signed it

blaggacao:
I signed it

blaggacao:
I signed it

blaggacao:
I signed it

deploy/hostpath/csi-hostpath-plugin.yaml
```diff
@@ -23,13 +23,15 @@ spec:
...
-            - --kubelet-registration-path=/var/lib/kubelet/plugins/csi-hostpath/csi.sock
+            - --kubelet-registration-path=$(DRIVER_REG_SOCK_DIR)/csi.sock
           env:
             - name: KUBE_NODE_NAME
               valueFrom:
                 fieldRef:
                   apiVersion: v1
                   fieldPath: spec.nodeName
+            - name: DRIVER_REG_SOCK_DIR
+              value: /opt/rke/var/lib/kubelet/plugins/csi-hostpath
```
pohly: This changes the `--kubelet-registration-path` for everyone to `/opt/rke/var/lib/kubelet/plugins/csi-hostpath`, which will be wrong for the majority of the potential users of this deployment.

What we need is a parameter for the `deploy.sh` script that overrides the default path, and keep the default the same as before.

blaggacao: Oh. My bad. Me copy pasting... sorry. Will change that back to the appropriate default.

blaggacao:
@pohly Putting a parameter into the `deploy.sh`, is there a way to keep this a `yaml` as oposed to a `yaml.tmpl`? Maybe some `sed` operation?

pohly:
Let's keep it a .yaml which has the "normal" value. We use it like that in E2E testing.

I think that a "sed" pre-processing step that replaces the "normal" value with the "actual" value in deploy.sh sounds good. Perhaps not as clean as some templating solution, but simpler.

I wonder what this would look like with kustomize support. I don't have personal experience with that.

blaggacao:
You inadvertently just explained to me the true value of `kustomize` :wink: 
It would probably look something like this:
````
deploy/base/...
deploy/overlays/rke/kustomization.yaml
deploy/overlays/rke/patch_plugin.yaml
deploy/overlays/.../patch_plugin.yaml
```
I'd personally prefer that over a `deploy.sh` because kustomize outputs the final deployment spec as yaml which I then can process further...

Though, the goal probably would not be to incorporate `deploy/overlays/XYZ` in here.

Would `kustomize` even replace `deploy.sh`?

pohly:
I suspect we have a few things in `deploy.sh` that cannot be done with `kustomize`, specifically determining the right version of the RBAC rules.

But it would be an interesting experiment to do the path parameterization with `kustomize`.

msau42:
> determining the right version of the RBAC rules.

What we do in the pd driver is to maintain two different kustomize overlays: one for stable that has fixed versions for everything, and one for latest.  There was some workarounds that we had to do for prow testing in order to substitute the image registry and tags, but I believe the latest release of kustomize has better support for this now.

Issue #12, opened 2019-02-22
Deployment by README steps fails in 4-node cluster


okartau:
Deployment as shown in README will succeed for sure only if attacher, provisioner, hostpathplugin and my-csi-app run in the same node. When I try the steps in 4-node cluster, the pods state will be:
```
NAME                         READY   STATUS              RESTARTS   AGE     IP            NODE   
csi-hostpath-attacher-0      1/1     Running             0          3m56s   10.244.1.59   host-1
csi-hostpath-provisioner-0   1/1     Running             0          3m56s   10.244.2.11   host-2 
csi-hostpathplugin-8knjt     2/2     Running             0          3m56s   192.168.8.4   host-1
csi-hostpathplugin-bw855     2/2     Running             0          3m56s   192.168.8.8   host-3
csi-hostpathplugin-dv8x7     2/2     Running             0          3m56s   192.168.8.6   host-2
my-csi-app                   0/1     ContainerCreating   0          3m3s    <none>        host-3
```

my-csi-app does not reach Running state as it can't mount:
```
mount: mounting /tmp/94131d40-35c8-11e9-946e-deadbeef0102 on
/var/lib/kubelet/pods/e884bdff-367e-11e9-8553-deadbeef0100/volumes/kubernetes.io~csi/pvc-8d479cfd-35c8-11e9-8553-deadbeef0100/mount
failed: No such file or directory
```

okartau:
@pohly proposed these changes to get example deployment working:
- start one instance of csi-hostpathplugin
- use inter-pod affinity to land attacher, provisioner, hostpathplugin and my-csi-app on same node

If we agree on this, I will prepare a PR

blaggacao:
I observe the same error:
```
Events:
  Type     Reason                  Age                From                     Message
  ----     ------                  ----               ----                     -------
  Warning  FailedScheduling        17s (x4 over 17s)  default-scheduler        pod has unbound immediate PersistentVolumeClaims (repeated 3 times)
  Normal   Scheduled               17s                default-scheduler        Successfully assigned gitops/atlantis-0 to 10.0.1.78
  Normal   SuccessfulAttachVolume  17s                attachdetach-controller  AttachVolume.Attach succeeded for volume "pvc-7b2a9f04-36ae-11e9-8852-52540082fa9b"
[...]

PR #14, opened 2019-02-22
Change example deployment scheme in multi-node cluster

okartau:
Current scheme does not work in multi-node cluster because
my-csi-app pod may land on different node than other pods
and mount point is not there.
This commit changes deployment so that there is one
instance of csi-hostpathplugin, and we use inter-pod affinity
to land attacher, provisioner, hostpathplugin and my-csi-appi
on a same node.

Fixes: #12

### DIFF
--- a/deploy/hostpath/csi-hostpath-plugin.yaml
+++ b/deploy/hostpath/csi-hostpath-plugin.yaml
@@ -1,8 +1,31 @@
-kind: DaemonSet
+# Service defined here, plus serviceName below in StatefulSet,
+# are needed only because of condition explained in
+# https://github.com/kubernetes/kubernetes/issues/69608
+
+kind: Service
+apiVersion: v1
+metadata:
+  name: csi-hostpathplugin
+  labels:
+    app: csi-hostpathplugin
+spec:
+  selector:
+    app: csi-hostpathplugin
+  ports:
+    - name: dummy
+      port: 12345
+---
+kind: StatefulSet
 apiVersion: apps/v1
 metadata:
   name: csi-hostpathplugin
 spec:
+  serviceName: "csi-hostpathplugin"
+  # One replica only:
+  # Host path driver only works when everything runs
+  # on a single node. We achieve that by starting it once and then
+  # co-locate all other pods via inter-pod affinity
+  replicas: 1
   selector:
     matchLabels:
       app: csi-hostpathplugin
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ rolebinding.rbac.authorization.k8s.io/csi-attacher-role-cfg created
 deploying hostpath components
 service/csi-hostpath-attacher created
 statefulset.apps/csi-hostpath-attacher created
-daemonset.apps/csi-hostpathplugin created
+statefulset.apps/csi-hostpathplugin created
 service/csi-hostpath-provisioner created
 statefulset.apps/csi-hostpath-provisioner created
 ```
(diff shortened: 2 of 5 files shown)


okartau:
This addresses #12. 
With these changes, I can see all related pods running in same node in 4-node cluster, and my-csi-app reaches Running state.

deploy/hostpath/csi-hostpath-plugin.yaml
```diff
@@ -1,8 +1,9 @@
-kind: DaemonSet
+kind: StatefulSet
 apiVersion: apps/v1
 metadata:
   name: csi-hostpathplugin
 spec:
+  serviceName: "csi-hostpathplugin"
```
pohly: I know it is a bit dumb in this particular usage of StatefulSet, but until someone from SIG-App declares that leaving out the Service is okay (see https://github.com/kubernetes/kubernetes/issues/69608) we should have a fake Service with that name defined in this .yaml file.

okartau:
I force-pushed, adding minor change in README.md where name changes slightly based on pod types we create.

deploy/hostpath/csi-hostpath-plugin.yaml
pohly: There's still no `csi-hostpathplugin` service, is it? Please add it to this file and link to that issue in the explanation why we have the useless `Service`.

okartau: added Service with comment, force-pushed

pohly:
/lgtm

Perhaps add `Fixes: #12` to the PR description, then that issue will get closed automatically on merge.

cwdsuzhou:
I do not think it is a good idea to use StatefulSet to deploy the driver. Though it can fix #12, it only deploy it on one node. We only need make the app, the external-provsioner and external-attacher run on the same node. In addition, since the hostpathdriver does not support `ControllerPublish`, we can use cluster-driver-registrar. It indicates to Kubernetes to skip the attach operation and eliminates the need to deploy the external-attacher sidecar. [https://kubernetes-csi.github.io/docs/cluster-driver-registrar.html#csi-cluster-driver-registrar](url)

pohly:
> We only need make the app, the external-provsioner and external-attacher run on the same node.

And then we have n copies of the driver, but only one is ever going to be used. How is that better?

> In addition, since the hostpathdriver does not support ControllerPublish, we can use cluster-driver-registrar.

That is a useful enhancement, but should be done separately.

cwdsuzhou:
> Current scheme does not work in multi-node cluster because
> my-csi-app pod may land on different node than other pods
> and mount point is not there.
> This commit changes deployment so that there is one
> instance of csi-hostpathplugin, and we use inter-pod affinity
> to land attacher, provisioner, hostpathplugin and my-csi-appi
> on a same node.
> 
> Fixes: #12



> > We only need make the app, the external-provsioner and external-attacher run on the same node.
> 
> And then we have n copies of the driver, but only one is ever going to be used. How is that better?
> 

Sorry, from the resource, it is better to use statefulSet to deploy it.
But the service is useless. 
Overall, the hostpathdriver is better to run in a single-node cluster

deploy/hostpath/csi-hostpath-plugin.yaml
vladimirvivien: Maybe add `replicas: 1` to be clear.

okartau:
> Maybe add `replicas: 1` to be clear.

added replicas: 1 with comment

PR #16, opened 2019-02-26
initial prow job

pohly:
This implements the kubernetes-csi release process described in https://github.com/kubernetes/enhancements/pull/807.

### DIFF
--- a/.prow.sh
+++ b/.prow.sh
@@ -1,0 +1,16 @@
+#! /bin/bash
+
+# TODO: move into job
+CSI_PROW_GINKO_PARALLEL="-p -nodes 40" # default was 7
+
+# Simulates canary test job.
+# TODO: canary periodic job
+#CSI_PROW_BUILD_JOB=false
+#CSI_PROW_KUBERNETES_VERSION=latest
+#CSI_PROW_HOSTPATH_CANARY=canary
+
+CSI_PROW_TESTS_SANITY="sanity"
+
+. release-tools/prow.sh
+
+main
--- a/release-tools/.prow.sh
+++ b/release-tools/.prow.sh
@@ -1,0 +1,7 @@
+#! /bin/bash -e
+#
+# This is for testing csi-release-tools itself in Prow. All other
+# repos use prow.sh for that, but as csi-release-tools isn't a normal
+# repo with some Go code in it, it has a custom Prow test script.
+
+./verify-shellcheck.sh "$(pwd)"
--- a/release-tools/README.md
+++ b/release-tools/README.md
@@ -49,3 +49,60 @@ Cheat sheet:
 - `git subtree add --prefix=release-tools https://github.com/kubernetes-csi/csi-release-tools.git master` - add release tools to a repo which does not have them yet (only once)
 - `git subtree pull --prefix=release-tools https://github.com/kubernetes-csi/csi-release-tools.git master` - update local copy to latest upstream (whenever upstream changes)
 - edit, `git commit`, `git subtree push --prefix=release-tools git@github.com:<user>/csi-release-tools.git <my-new-or-existing-branch>` - push to a new branch before submitting a PR
+
+verify-shellcheck.sh
+--------------------
+
+The [verify-shellcheck.sh](./verify-shellcheck.sh) script in this repo
+is a stripped down copy of the [corresponding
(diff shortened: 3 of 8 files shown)


pohly:
/test pull-sig-storage-csi-driver-host-path

Point to remember: all kubernetes-csi repos must be listed in prow/plugins.yaml before commands work, see https://github.com/kubernetes/test-infra/pull/11537.

BenTheElder:
```console
+ cd /home/prow/go/src/k8s.io/kubernetes
+ ginkgo -p -v ./test/e2e -- -ginkgo.focus=External.Storage.*csi-hostpath -ginkgo.skip=\[Feature:|\[Disruptive\] -storage.testdriver=/home/prow/go/pkg/csiprow.kor3UsM3eJ/hostpath-test-driver.yaml
./.prow.sh: 244: ./.prow.sh: ginkgo: not found
```
the kind part works fine now, the script needs some fixing :+1:

pohly:
First job where E2E tests ran: https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_csi-driver-host-path/16/pull-sig-storage-csi-driver-host-path/14

There's one E2E failure (`External Storage [Driver: csi-hostpath] [Testpattern: Dynamic PV (block volmode)] volumeMode [It] should fail in binding dynamic provisioned PV to PVC`). I need to check whether that is something that the hostpath driver is supposed to  pass.

BenTheElder:
FYI currently junit output from ginkgo is not being picked up, if you make sure it is written to  somewhere in $ARTIFACTS Prow will render the structured results and let you view the failure output specifically.

pohly:
Benjamin Elder <notifications@github.com> writes:
> FYI currently junit output from ginkgo is not being picked up, if you
> make sure it is written to  somewhere in $ARTIFACTS Prow will render
> the structured results and let you view the failure output
> specifically.

I know, that's one of the TODOs.

Ginkgo is going to be easy. More interesting will be to also generate
JUnit output for the individual tests done by "make test". I'm currently
leaning towards leaving "make test" as-is and just parsing the output.

pohly:
The test failure was caused by https://github.com/kubernetes-csi/csi-driver-host-path/issues/21

BenTheElder:
I would recommend getting the ginkgo output FWIW, it should be pretty easy and make it quicker to diagnose failures from those tests. It just needs to be placed in $ARTIFACTS for upload.

/test pull-sig-storage-csi-driver-host-path

pohly:
@BenTheElder `make test` doesn't run Ginkgo tests, just various checks like gofmt.

/test pull-sig-storage-csi-driver-host-path

pohly:
@msau42 if the tests succeed, then please merge. This is up-to-date with https://github.com/kubernetes-csi/csi-release-tools/pull/9 (= d87eccb).

msau42:
Can you explain what these jobs are doing:
pull-sig-storage-csi-driver-host-path-kubernetes-1-13 
pull-sig-storage-csi-driver-host-path-kubernetes-master

Can we also separate out the non-K8s tests, alpha e2es and non-alpha e2es?

pohly:
> Can you explain what these jobs are doing:
> pull-sig-storage-csi-driver-host-path-kubernetes-1-13
> pull-sig-storage-csi-driver-host-path-kubernetes-master

Those are from previous revisions of test-infra and obsolete. The current set of tests are the ones from https://github.com/kubernetes-csi/csi-driver-host-path/pull/16#issuecomment-480057783

> Can we also separate out the non-K8s tests, alpha e2es and non-alpha e2es?

Yes, but that will be done in test-infra without requiring changes in the individual components.

pohly:
> > Can we also separate out the non-K8s tests, alpha e2es and non-alpha e2es?
>
>Yes, but that will be done in test-infra without requiring changes in the individual components.

See https://github.com/kubernetes/test-infra/pull/12088

While testing I noticed that sig-release-tools needs some enhancement and one fix to make this actually work (the check for which tests to run is slightly broken). I have the fixed, but would like to get this PR merged first, to unblock testing of the periodic jobs. Right now they aren't passing because they are based on csi-driver-host-path master, which doesn't have Prow support yet.

pohly:
@msau42 you or someone else (@jsafrane?) also needs to approve.

PR #20, opened 2019-03-01
Make provisioner root persistent across reboot

peter-wangxu:
If the proversioner container or the host restart
data dir under /tmp in the container will be lost thus
causing start up failure for any pods using csi pv.

It's good to make the provisioner root persistent so that
this plugin is  more reliable for daily testing.

### DIFF
--- a/deploy/hostpath/csi-hostpath-plugin.yaml
+++ b/deploy/hostpath/csi-hostpath-plugin.yaml
@@ -60,6 +60,8 @@ spec:
             name: socket-dir
           - mountPath: /registration
             name: registration-dir
+          - mountPath: /csi-data-dir
+            name: csi-data-dir
 
         - name: hostpath
           image: quay.io/k8scsi/hostpathplugin:v1.0.1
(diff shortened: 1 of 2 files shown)


deploy/hostpath/csi-hostpath-plugin.yaml
```diff
@@ -72,3 +74,7 @@ spec:
             path: /var/lib/kubelet/plugins_registry
             type: Directory
           name: registration-dir
+        - hostPath:
+            path: /var/lib/csi-hostpath-data/
```
jsafrane: I personally like /tmp because I don't need to worry about cleaning up failed tests or my debugging env.

For me it would be enough if there is a comment that this is path where data is persisted on the host & any directory such as /tmp can be used.

Any other opinion from @kubernetes-csi/csi-misc?

peter-wangxu: DONE

peter-wangxu:
@jsafrane any more comments?

jsafrane:
No one has raised any other comments
/lgtm
/approve

Issue #21, opened 2019-03-01
master not a complete copy of the original code


pohly:
When moving the hostpath driver, some changes were lost. It looks like an older version of the original code was copied, in particular this change here is missing: 6820279e588166dfa0e0013f6669b1a090fa1a96

git diff 6820279e588166dfa0e0013f6669b1a090fa1a96..660c3571d2389573b411e0898e02c85f7bcab28a pkg/hostpath/controllerserver.go
```
...
-       for _, cap := range caps {
-               if cap.GetBlock() != nil {
-                       return nil, status.Error(codes.Unimplemented, "Block Volume not supported")
-               }
-       }
...
```

The commit is there because the full history was spliced in, but the actual tree that it was spliced into was based on an older copy.

/assign

pohly:
Comparing between old and imported code with git diff c1e71bd..660c3571d2389573b411e0898e02c85f7bcab28a shows that this is the only lost code. The rest of the diff is what one would expect (old code removed, import paths changed, etc.).

pohly:
This got resolved when replacing the out-dated code with a new implementation (https://github.com/kubernetes-csi/csi-driver-host-path/pull/6).

PR #22, opened 2019-03-04
sync up master with "drivers"

pohly:
When copying from the "drivers" repo, one commit was missed. This is a cherry-pick of that commit.

Fixes: #21


msau42:
This may not be needed anymore after https://github.com/kubernetes-csi/csi-driver-host-path/pull/6

pohly:
> This may not be needed anymore after #6.

True.

PR #25, opened 2019-03-06
remove block file when losetup fails at volume creation

### LINKED ISSUE #23: Driver leaks files on creation failures
When creating a raw block volume and the losetup call fails, the created file is not cleaned up, and it consumes space in /tmp forever.

darkowlzz:
This change logs the attach device failure while creating a block
volume and tries to remove the associated block file which will not
be reused.

Fixes #23

### DIFF
--- a/pkg/hostpath/controllerserver.go
+++ b/pkg/hostpath/controllerserver.go
@@ -153,6 +153,11 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 		volPathHandler := volumepathhandler.VolumePathHandler{}
 		_, err = volPathHandler.AttachFileDevice(path)
 		if err != nil {
+			glog.Errorf("failed to attach device: %v", err)
+			// Remove the block file because it'll no longer be used again.
+			if err2 := os.Remove(path); err != nil {
+				glog.Errorf("failed to cleanup block file %s: %v", path, err2)
+			}
 			return nil, status.Error(codes.Internal, fmt.Sprintf("failed to attach device: %v", err))
 		}
 	case mountAccess:


pkg/hostpath/controllerserver.go
```diff
@@ -153,6 +153,11 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 		volPathHandler := volumepathhandler.VolumePathHandler{}
 		_, err = volPathHandler.AttachFileDevice(path)
 		if err != nil {
+			glog.V(3).Infof("failed to attach device: %v", err)
+			// Remove the block file because it'll no longer be used again.
+			if err := os.Remove(path); err != nil {
+				glog.V(3).Infof("failed to cleanup block file %s: %v", path, err)
```
bswartz: Log this at warn/error not info.

pkg/hostpath/controllerserver.go
```diff
@@ -153,6 +153,11 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 		volPathHandler := volumepathhandler.VolumePathHandler{}
 		_, err = volPathHandler.AttachFileDevice(path)
 		if err != nil {
+			glog.V(3).Infof("failed to attach device: %v", err)
```
bswartz: Log this at warn/error not info.

pkg/hostpath/controllerserver.go
```diff
@@ -153,6 +153,11 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
...
+			// Remove the block file because it'll no longer be used again.
+			if err := os.Remove(path); err != nil {
```
bswartz: This redeclartion of the "err" variable shadows the one in the outer scope and will cause the actual error returned to the caller to contain a nil value when losetup fails and the remove succeeds. I suggest using "err2" to avoid this.

darkowlzz: Variables within a conditional blocks are usually scoped within that block only.
But it's better to rename it to avoid any potential bugs due to duplicate variable names.
Thanks! :)

darkowlzz:
Added the suggested changes. Hope it's ready now.

pkg/hostpath/controllerserver.go
bswartz: Doh! I keep forgetting that nested scope starts at the if statement and not the curly brace. Good point.

PR #26, opened 2019-03-08
deploy: fix PR #20

pohly:
PR #20 introduced broken .yaml, causing a failure from kubectl:
   error: error parsing STDIN: error converting YAML to JSON: yaml: line 91: did not find expected key

### DIFF
--- a/deploy/hostpath/csi-hostpath-plugin.yaml
+++ b/deploy/hostpath/csi-hostpath-plugin.yaml
@@ -106,6 +106,7 @@ spec:
             path: /var/lib/kubelet/plugins
             type: Directory
           name: plugins-dir
+        - hostPath:
             # 'path' is where PV data is persisted on host.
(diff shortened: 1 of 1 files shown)


deploy/hostpath/csi-hostpath-plugin.yaml
```diff
@@ -105,7 +105,6 @@ spec:
         - hostPath:
             path: /var/lib/kubelet/plugins
             type: Directory
-          name: plugins-dir
```
msau42: i don't think this is correct? we need plugins-dir defined.  I think the next entry afterwards needs a 
```
-hostPath
```

pohly: Now that you mention it, yes.

pohly: @msau42 Force-pushed.

PR #27, opened 2019-03-10
Added liveness probe to deployment example manifest

okartau:
Liveness probe configuration is added based on deployment
example in kubernetes-csi/livenessprobe.
Different port value is used than default, for showing
how it's configured.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,8 @@ but as this is something that should be done by the cluster
 provisioning tool it is disabled in the script by default. For this
 and other customizations see the source code of the deploy script.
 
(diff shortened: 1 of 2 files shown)


okartau:
I added a sentence in README mentioning liveness probe (combining the meaning currently expressed in few sentences in doc example that we are about to reduce).
What do you think, should we mark the liveness-related parts in yaml file with specific comment?
Without such marking, the liveness part dissolves into the whole and someone who might want to deploy without liveness checker, has to study which part to cut.

okartau:
I have kept a sample deployment instance running with this change included for last 16 days.
It has been just idle, but at least liveness-probe part has got something to do.
No problems recorded, no restarts.
Can we get this merged to get https://github.com/kubernetes-csi/docs/pull/125 moving ?

PR #29, opened 2019-03-27
deployment enhancements

pohly:
This was done for the Kubernetes-CSI CI work (PR #16), but as these are useful enhancements by themselves and perhaps even required before we can release something for Kubernetes 1.14 I suggest to review and merge them  separately.

This PR is best reviewed commit-by-commit.


deploy/kubernetes-1.13/README.md
```diff
@@ -0,0 +1,9 @@
+The deployment for Kubernetes 1.13 uses CSI 1.0 and this is
+incompatible with older Kubernetes releases.
+
+It relies on the CRDs for CSIDriverInfo and CSINodeInfo, which are
+about to be replaced with builtin APIs in Kubernetes 1.14. It can be
+deployed on Kubernetes 1.14 if the CRDs are installed, but features
+relying on these CRDs (like topology) are unlikely to work.
+
+Kubernetes 1.14 will need a different deployment.
```
pohly: So what are we going to do for Kubernetes 1.14?

Some options:
1. introduce a kubernetes-1.14 deployment with the upcoming sidecars and all features enabled (like topology) -> that explains why a second deployment is needed because there's some real difference
2. introduce a kubernetes-1.14 deployment with the upcoming sidecars and no features enabled (topology off) -> the only difference then is the image versions, both deployments will work on Kubernetes 1.13 and 1.14
3. bump the image versions in `kubernetes-1.13` and symlink `kubernetes-1.14` to it -> everyone will use the latest hostpath driver, but we no longer have an example using the CRD-based sidecars

Whatever we do will have implications for CI testing. But let's ignore that here and decide based on what is best for users (= people who want to try out CSI and/or need a working example).

/cc @msau42

msau42: I think we should have different directories for each supported kubernetes release.

Hostpath doesn't use any of the crd based features so I'm not sure if it makes sense to include them because they won't actually get tested with those configurations.

pohly: So that's option 2 then, right?

pohly: I've pushed an update with `deploy/kubernetes-1.14`, using the canary images as interim solution.

deploy/kubernetes-1.14/hostpath/csi-hostpath-attacher.yaml
```diff
@@ -0,0 +1,55 @@
...
+    app: csi-hostpath-attacher
+  ports:
+    - name: dummy
+      port: 12345
+
+---
+kind: StatefulSet
+apiVersion: apps/v1
+metadata:
+  name: csi-hostpath-attacher
```
msau42: Is there a reason why we need to have each controller be its own Pod?  Can we combine all the controller sidecars and driver into one controller deployment since that is what we would encourage driver vendors to do?

deploy/kubernetes-1.13/README.md
```diff
@@ -0,0 +1,7 @@
+The deployment for Kubernetes 1.13 uses CSI 1.0 and thus is
+incompatible with older Kubernetes releases.
+
+The sidecar images rely on the CRDs for CSIDriverInfo and CSINodeInfo,
```
msau42: Should we have different directories for alpha features? This makes it easier to test forward compatibility of a sidecar using only beta/ga features

deploy/deploy-hostpath.sh
```diff
@@ -11,38 +11,106 @@
...
+    echo "$version"
 }
 
-BASE_DIR=$(dirname "$0")
-K8S_RELEASE=${K8S_RELEASE:-"release-1.13"}
-PROVISIONER_RELEASE=${PROVISIONER_RELEASE:-$(image_version "${BASE_DIR}/hostpath/csi-hostpath-provisioner.yaml" csi-provisioner)}
-ATTACHER_RELEASE=${ATTACHER_RELEASE:-$(image_version "${BASE_DIR}/hostpath/csi-hostpath-attacher.yaml" csi-attacher)}
-SNAPSHOTTER_RELEASE=${SNAPSHOTTER_RELEASE:-$(image_version "${BASE_DIR}/snapshotter/csi-hostpath-snpshotter.yaml" csi-snapshotter)}
+# In addition, the RBAC rules can be overridden for provisioner and attacher.
+CSI_PROVISIONER_RBAC=${PROVISIONER_RBAC:-https://raw.githubusercontent.com/kubernetes-csi/external-provisioner/$(image_version "${BASE_DIR}/hostpath/csi-hostpath-provisioner.yaml" csi-provisioner)/deploy/kubernetes/rbac.yaml}
```
pohly: The idea here is that for a PR against external-provisioner, we always test with the RBAC rules that match the code of the external-provisioner that is getting tested. That way modified RBAC rules get tested.

When testing "canary" images, we use the existing deployment, but with updated image and updated RBAC.

However, in the versioning doc we recently agreed to define that updated sidecars also must be compatible with the RBAC in older deployments, which means we need to test *also* with the original RBAC.

This further multiplies our testing matrix :frowning: 

What about this:
- in a PR, we use the potentially modified RBAC rules
- the canary test jobs get defined in two flavors, with replaced RBAC and without

deploy/kubernetes-1.13/README.md
pohly: Perhaps the comment is misleading: this particular example does specify anything related to alpha features and thus can be used unchanged on Kubernetes 1.13 and 1.14.

What I was trying to say here is that if someone uses the same sidecars as the example and depends on alpha features like topology, then his deployment won't work on Kubernetes 1.14.

We may need an "alpha" variant of these examples at some point (for example, after adding topology support to the hostpath driver), but right now I think they are not needed.

deploy/kubernetes-1.13/hostpath/csi-hostpath-service.yaml
```diff
@@ -0,0 +1,56 @@
...
+apiVersion: v1
+kind: Service
+metadata:
+  name: hostpath-service
+spec:
+  type: NodePort
+  selector:
+    app: csi-hostpath-socat
+  ports:
+  - port: 10000 # fixed port inside the pod, dynamically allocated port outside
```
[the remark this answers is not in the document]
pohly: Okay.

deploy/kubernetes-1.14/README.md
```diff
@@ -0,0 +1,10 @@
+The deployment for Kubernetes 1.14 uses CSI 1.0 and thus is incompatible with
+Kubernetes < 1.13.
+
+It uses the builtin APIs for CSIDriverInfo and CSINodeInfo that were
+introduced in Kubernetes 1.14, so features depending on those (like
+topology) will not work on Kubernetes 1.13. But because this example
+deployment does not enable those features, it can run on Kubernetes 1.13.
+
+WARNING: the images for Kubernetes 1.14 have not been released yet, so this
```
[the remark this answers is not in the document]
pohly: Some of the CI jobs expect to find a "kubernetes-1.14" when testing on that version.

I'll rename the directory to "master" and add a symlink, that should make it obvious that this is not the final deployment for 1.14.

deploy/kubernetes-1.14/hostpath/csi-hostpath-attacher.yaml
```diff
@@ -0,0 +1,55 @@
+kind: Service
```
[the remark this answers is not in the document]
pohly: I never got a confirmation from SIG-apps that it really isn't required. See https://github.com/kubernetes/kubernetes/issues/69608

I agree, we should switch to a Deployment. But let's do that separately.

deploy/kubernetes-1.14/hostpath/csi-hostpath-attacher.yaml
pohly: The reason is that with separate pods we can verify that the RBAC rules for that sidecar are really accurate.

If we were to combine them, we would have to create a new service account which has all permissions of all sidecars. This is doable, but then we don't catch mistakes like "external-attacher needs a permission that is only defined in the RBAC rules for external-provisioner".

From a security perspective, running separate pods is good ("least privilege" principle).

Having said that, personally I would prefer to release just a single binary for the controller operations. I don't think the alleged advantages of one repo/one binary for each of the sidecars (more agile development, independent versioning and releases) are worth the confusion that this creates for us (as developers of the sidecars) and the users (who now need to track four, no, five! components instead of just one).

deploy/deploy-hostpath.sh
pohly: I've pushed an updated version where the caller can control whether the RBAC rules are used unchanged or get updated together with the image.

How we want to use that can be discussed in more detail in https://github.com/kubernetes/test-infra/pull/11981

deploy/kubernetes-1.13/README.md
```diff
@@ -0,0 +1,9 @@
+The deployment for Kubernetes 1.13 uses CSI 1.0 and this is
+incompatible with older Kubernetes releases.
+
+It relies on the CRDs for CSIDriverInfo and CSINodeInfo, which are
```
msau42: I guess this is confusing because the hostpath driver doesn't actually use these features.  Instead, should we have a K8s 1.14 changelog page on our docs page that describes all the new kubernetes changes?

pohly: Yes, that would be better and can be more comprehensive. We still should summarize here, though.

deploy/master/deploy-hostpath.sh
```diff
@@ -0,0 +1,162 @@
...
+# done here and referenced elsewhere.
+
+# The script assumes that kubectl is available on the OS path 
+# where it is executed.
+
+set -e
+set -o pipefail
+
+BASE_DIR=$(dirname "$0")
+K8S_RELEASE=${K8S_RELEASE:-"release-1.13"}
```
msau42: This should be master?

deploy/master/deploy-hostpath.sh
```diff
@@ -0,0 +1,162 @@
...
+
+# If set, the following env variables override image registry and/or tag for each of the images.
```
msau42: Is there any difference in the rest of this script compared to the 1.13 version?  If not, then maybe this should be a common script shared by both 1.13 and master deployments.  The only difference is the K8s and image versions used.

deploy/master/deploy-hostpath.sh
pohly: No, this is used further down for installing the CRDs. I think those should come from the branch which actually uses them.

But this code is pretty much dead. It's not used unless explicitly requested. @vladimirvivien added it at a time when there was uncertainty around who's responsible for installing the CRDs. Should we perhaps remove it entirely, in both versions of the script?

deploy/master/deploy-hostpath.sh
pohly: It is identical. The image versions are not hard-coded here either but rather taken from the yaml files.

The reason why I kept them separate is that they might eventually become different, so the CI plays it safe and uses the version of the script that it finds in the deployment directory.

We could avoid the duplication by
- moving the script to an "util" directory and symlinking to it
- keeping the file in kubernetes-1.13 and symlink to it from kubernetes-1.14

I don't like the second option. It'll be too easy to make a change without realizing that it'll affect other deployments.

deploy/master/deploy-hostpath.sh
msau42: Yeah let's remove it entirely.

deploy/master/deploy-hostpath.sh
msau42: I prefer the first option too.

pohly: Done.

deploy/master/deploy-hostpath.sh
pohly: Done.

README.md
```diff
@@ -8,46 +8,62 @@ This repository hosts the CSI Hostpath driver and all of its build and dependent
 - Access to terminal with `kubectl` installed
 
 ## Deployment
-The easiest way to test the Hostpath driver is to run `deploy/deploy-hostpath.sh` script as shown:
+The easiest way to test the Hostpath driver is to run the `deploy-hostpath.sh` script for the Kubernetes version used by
+the cluster as shown below for Kubernetes 1.13. This creates the deployment that is maintained specifically for that
+release of Kubernetes. However, other deployments may also work. For details see the individual READMEs.
 
 ```shell
-$ deploy/deploy-hostpath.sh
```
msau42: Why not master? So we don't have to keep on updating this readme whenever a new kubernetes release comes out

pohly:
@msau42 force-pushed a rebased version with the discussed changes. README.md gets updated together with the script now.

deploy/util/deploy-hostpath.sh
```diff
@@ -0,0 +1,166 @@
...
+# done here and referenced elsewhere.
+
+# The script assumes that kubectl is available on the OS path 
+# where it is executed.
+
+set -e
+set -o pipefail
+
+BASE_DIR=$(dirname "$0")
+K8S_RELEASE=${K8S_RELEASE:-"release-1.13"}
```
msau42: Can you rename this as CSI_CRD_INSTALL_RELEASE to make it more obvious what this is used for?

msau42:
/lgtm
/approve

Minor comments can be followed up separately.

pohly:
> Can you rename this as CSI_CRD_INSTALL_RELEASE to make it more obvious what this is used for?

Actually, I wanted to remove the CRD install code and had done so, but then resolved a merged conflict incorrectly and it came back. 

Will remove it in a follow-up PR.

PR #31, opened 2019-03-29
Add snapshot deployment example

okartau:
Migrated the snapshot example from docs/example to here,
with minor updates and cleanup.
Corrected misspelled snpshot in deploy file name.


README.md
```diff
@@ -14,7 +14,7 @@ The easiest way to test the Hostpath driver is to run `deploy/deploy-hostpath.sh
 $ sh deploy/deploy-hostpath.sh
 ```
 
-You should see an output similar to the following printed on the terminal showing the application of rbac rules and the result of deploying the hostpath driver, external privisioner and external attacher components:
+You should see an output similar to the following printed on the terminal showing the application of rbac rules and the result of deploying the hostpath driver, external privisioner, external attacher and snapshotter components:
```
pohly: Chance for a drive-by spelling fix:
s/privisioner/provisioner/

README.md
```diff
@@ -51,7 +52,7 @@ The [livenessprobe side-container](https://github.com/kubernetes-csi/livenesspro
 
 ## Run example application and validate
 
-Next, validate the deployment.  First, ensure all expected pods are running properly including the external attacher, provisioner, and the actual hostpath driver plugin:
+Next, validate the deployment.  First, ensure all expected pods are running properly including the external attacher, provisioner, snapshotter and the actual hostpath driver plugin:
```
pohly: The following output probably also needs to be updated?

README.md
```diff
@@ -188,6 +189,141 @@ Events:      <none>
 ```
 
 
+## Snapshot support
+
+Since volume snapshot is an alpha feature in Kubernetes v1.12 and v1.13, you need to enable feature gate called `VolumeSnapshotDataSource` in the Kubernetes API server binary.
```
pohly: @xing-yang it's also alpha in 1.14, right?

README.md
okartau: yes, those 4 spelling errors sting my eye every time, but there is #2 to fix those and I did not want to step on toes of another ongoing change. I can sure fix in current PR if so desired.

README.md
okartau: no, it was updated with one of previous PRs (added snapshotter pod). In current PR I noticed the list in text lacks the snapshotter

README.md
pohly: Please include the fix. Otherwise we just end up with merge conflicts.

okartau: yes, makes sense, as this PR makes massive changes anyway. And I proposed few of #2 fixes there, so I hope its not that intrusive. Will change and re-push. 
BTW, while we discuss spell etc minor issues, I want to propose: lets get rid of explicit shell call in deploy script start cmd. As the script has x-bits set, it can be started without specifying shell explicitly, and I dont see extra value from explicit sh, is there?

okartau: I included all fixes of #2 now, plus removed explicit sh in deploy cmd, force-pushed

pohly: > lets get rid of explicit shell call in deploy script start cmd

You mean `#!/usr/bin/env bash`?

> As the script has x-bits set, it can be started without specifying shell explicitly, 

Leave out the shebang line entirely? I doubt that this is POSIX-compliant. For example, does it work because the shell itself then reads the input or is it the Linux kernel which falls back to running in a shell? A quick test (create executable file with `sleep 120`, invoke in script in dash and bash, look at `/proc/*/maps` of the script) shows that it is a feature of the shell: the script then always runs in the same shell as the one it was invoked in - if the shell supports that.

> and I dont see extra value from explicit sh, is there?

Last but not least, it does matter. The script uses bash features (`set -o pipefail`) and therefore bash has to be set explicitly.

pohly: >  plus removed explicit sh in deploy cmd

I'm not seeing that in the latest changes. Are you sure you force-pushed?

okartau: I was not clear enough in my comments. I did not mean deploy script contents, but how it's activation is written in README (where user typically copies commands from). 
Was: `sh deploy/deploy-hostpath.sh`
I changed to: `deploy/deploy-hostpath.sh`

pohly: Oh, absolutely. Invoking it in `sh` (which could be `dash`) is just plain wrong.

README.md
msau42: It also needs to be enabled in kube-controller-manager.

msau42: (and yes, alpha in 1.14)

okartau: I will add v1.14

msau42: Also regarding the alpha versions, maybe just saying "starting in Kubernetes v1.12" is good enough. Then we don't have to keep on updating the the example every Kubernetes release.

okartau: pushed new state with changes discussed here in recent comments

README.md
```diff
@@ -188,6 +189,143 @@ Events:      <none>
 ```
 
 
+## Snapshot support
+
+Since volume snapshot is an alpha feature starting in Kubernetes v1.12, you need to enable feature gate called `VolumeSnapshotDataSource` in the Kubernetes.
```
msau42: nit: reword as

"Volume snapshot is an alpha feature introduced in Kubernetes v1.12.  The Kubernetes feature gate `VolumeSnapshotDataSource` must be enabled."

pohly:
/hold

The script has a race condition: the snapshot class gets created without waiting for the snapshotter to create the corresponding CRD.

I suggest we merge this as part of PR #29 with a fix.

PR #34, opened 2019-04-07
deploy: avoid replacing socat image

pohly:
Setting IMAGE_TAG or IMAGE_REGISTRY unintentionally also affected the
socat image. This also disables SOCAT_IMAGE_TAG/REGISTRY, but that
should not be needed and wasn't documented either.

### DIFF
--- a/deploy/util/deploy-hostpath.sh
+++ b/deploy/util/deploy-hostpath.sh
@@ -164,4 +164,4 @@ done
 
 # deploy snapshotclass
 echo "deploying snapshotclass"
-kubectl create -f ${BASE_DIR}/snapshotter/csi-hostpath-snapshotclass.yaml
+kubectl apply -f ${BASE_DIR}/snapshotter/csi-hostpath-snapshotclass.yaml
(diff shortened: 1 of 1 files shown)


pohly:
/hold

Still needs to be tested.

pohly:
/hold cancel

Here's a succesful test run with this change: https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-csi-canary-on-kubernetes-master/1114802533932470277

deploy/util/deploy-hostpath.sh
```diff
@@ -105,7 +105,7 @@ for i in $(ls ${BASE_DIR}/hostpath/*.yaml | sort); do
     echo "   $i"
     modified="$(cat "$i" | while IFS= read -r line; do
         nocomments="$(echo "$line" | sed -e 's/ *#.*$//')"
-        if echo "$nocomments" | grep -q '^\s*image:\s*'; then
+        if echo "$nocomments" | grep -q '^\s*image:\s*' && ! echo "$nocomments" | grep -q image:.*/socat; then
```
msau42: Can we use a variable for blacklisted components?

PR #35, opened 2019-04-08
replace '\s' to [[:space:]] for supporting macos

pengzhisun:
I've tried to run the deploy script in MacOS but got following error message:
> error: error parsing STDIN: error converting YAML to JSON: yaml: line 30: could not find expected ':'

It seems the root cause for this error is the `sed` command in MacOS not supports '\s', so the modified yaml file contains invalid image statements like "image:quay.io/k8scsi/csi-attacher:canary"(there should be a space after colon).

The solution was referenced from [here](https://stackoverflow.com/questions/18840175/find-and-replace-with-spaces-using-sed-mac-terminal):
> For POSIX compliance, use the character class [[:space:]] instead of \s, since the latter is a GNU sed extension.


pengzhisun:
I signed it

PR #37, opened 2019-04-11
deploy: remove imagePullPolicy for livenessprobe

pohly:
Just as for the other containers the image is immutable and therefore
doesn't have to be checked all the time. Required for CI testing,
because it uses an image that is only available locally.


pohly:
This will fix the test failures in https://github.com/kubernetes-csi/livenessprobe/pull/41

PR #42, opened 2019-04-15
Fixed the Raw Block PV snapshot support

### LINKED ISSUE #41: Raw block volume snapshot failure
Raw block volume snapshot fails with error like:
```
 Warning  SnapshotCreationFailed  6s    csi-snapshotter csi-hostpath  Failed to create snapshot: failed to take snapshot of the volume, pvc-cb8d96b6-5d40-11e9-bcc0-0236f32917fe: "rpc error: code = Internal desc = failed create snapshot: exit status 1: tar: can't change directory to '/csi-data-dire6fc6eb8-5d40-11e9-97b2-0236f32917fe': Not a directory\n"
```
There are 2 reasons to this:
1 : Following constants are defined incorrect:
```
provisionRoot      = "/csi-data-dir"
snapshotRoot       = "/csi-data-dir"
```
Thjey should be 
```
provisionRoot      = "/csi-data-dir/"
snapshotRoot       = "/csi-data-dir/"
```
2: In case of raw block volumes the volume content comes as a single file as against a directory in case of a filesystem volume. So the tar command with -C option won't work. Instead, a tar file should be created directly from the file corresponding to the raw block volume. Something like
``` 
tar czf <tar_fikle_name> <file_for_raw_block_volume>
```

aayushrangwala:
/kind bug

Problem:
When Try to create the snapshot of a Raw Block PV, it fails with the following error, if volumesnapshot is described:
```
 Warning  SnapshotCreationFailed  6s    csi-snapshotter csi-hostpath  Failed to create snapshot: failed to take snapshot of the volume, pvc-cb8d96b6-5d40-11e9-bcc0-0236f32917fe: "rpc error: code = Internal desc = failed create snapshot: exit status 1: tar: can't change directory to '/csi-data-dire6fc6eb8-5d40-11e9-97b2-0236f32917fe': Not a directory\n"
```

The issue is described in [https://github.com/kubernetes-csi/csi-driver-host-path/issues/41](url)

Resolution:
1. Corrected the directory format by adding a "/"
2. The "tar" command in case of Block volume should be different. It should consider the vol as a file, which in case of filesystem Mode PV, it is considered as directory
3. The menifests in ./deploy/hostpath is updated, as the hostpathplugin container should be mounted to the "/csi-snapshot-dir" 

Fixes #41 

NONE

### DIFF
--- a/deploy/kubernetes-1.13/hostpath/csi-hostpath-plugin.yaml
+++ b/deploy/kubernetes-1.13/hostpath/csi-hostpath-plugin.yaml
@@ -37,7 +37,7 @@ spec:
       hostNetwork: true
       containers:
         - name: node-driver-registrar
-          image: quay.io/k8scsi/csi-node-driver-registrar:v1.0.2
+          image: quay.io/k8scsi/csi-node-driver-registrar:canary
           lifecycle:
             preStop:
               exec:
@@ -63,7 +63,7 @@ spec:
             name: csi-data-dir
 
         - name: hostpath
-          image: quay.io/k8scsi/hostpathplugin:v1.1.0-rc1
+          image: quay.io/k8scsi/hostpathplugin:canary
           args:
             - "--v=5"
             - "--endpoint=$(CSI_ENDPOINT)"
--- a/deploy/kubernetes-1.13/hostpath/csi-hostpath-plugin.yaml
+++ b/deploy/kubernetes-1.13/hostpath/csi-hostpath-plugin.yaml
@@ -99,6 +99,8 @@ spec:
             - mountPath: /var/lib/kubelet/plugins
               mountPropagation: Bidirectional
               name: plugins-dir
+            - mountPath: /csi-data-dir
+              name: csi-data-dir
 
         - name: liveness-probe
           volumeMounts:
--- a/deploy/kubernetes-1.13/hostpath/csi-hostpath-plugin.yaml
+++ b/deploy/kubernetes-1.13/hostpath/csi-hostpath-plugin.yaml
@@ -37,7 +37,7 @@ spec:
       hostNetwork: true
       containers:
         - name: node-driver-registrar
-          image: quay.io/k8scsi/csi-node-driver-registrar:canary
+          image: quay.io/k8scsi/csi-node-driver-registrar:v1.0.2
           lifecycle:
             preStop:
               exec:
--- a/pkg/hostpath/controllerserver.go
+++ b/pkg/hostpath/controllerserver.go
@@ -38,8 +38,8 @@ import (
 
 const (
 	deviceID           = "deviceID"
-	provisionRoot      = "/csi-data-dir"
-	snapshotRoot       = "/csi-data-dir"
+	provisionRoot      = "/csi-data-dir/"
+	snapshotRoot       = "/csi-data-dir/"
 	maxStorageCapacity = tib
 )
 
@@ -347,7 +347,14 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 	creationTime := ptypes.TimestampNow()
 	volPath := hostPathVolume.VolPath
 	file := snapshotRoot + snapshotID + ".tgz"
-	args := []string{"czf", file, "-C", volPath, "."}
+	args := []string{}
+	if hostPathVolume.VolAccessType == blockAccess {
+		glog.V(4).Infof("Creating snapshot of Raw Block Mode Volume")
+		args = []string{"czf", file, volPath}
+	} else {
+		glog.V(4).Infof("Creating snapshot of Filsystem Mode Volume")
+		args = []string{"czf", file, "-C", volPath, "."}
+	}
 	executor := utilexec.New()
 	out, err := executor.Command("tar", args...).CombinedOutput()
 	if err != nil {
--- a/pkg/hostpath/controllerserver.go
+++ b/pkg/hostpath/controllerserver.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"sort"
 	"strconv"
+	"strings"
 
 	"github.com/golang/protobuf/ptypes"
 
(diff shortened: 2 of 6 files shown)


aayushrangwala:
Signed the CLA

xing-yang:
Can you add an example yaml file to create a PVC with raw block mode and create a snapshot from it?

aayushrangwala:
> Can you add an example yaml file to create a PVC with raw block mode and create a snapshot from it?

Added 3 files, one creates a PVC which will create a Block mode PV, csi-pod-raw.yaml creates a pod to consume the block pv and csi-raw-pv-snapshot.yaml creates the snapshot

Pre-requisites:
csi-hostpath-snapshotclass.yaml to be deployed from /deploy/master/snapshotter

deploy/kubernetes-1.13/hostpath/csi-hostpath-plugin.yaml
```diff
@@ -99,6 +99,8 @@ spec:
             - mountPath: /var/lib/kubelet/plugins
               mountPropagation: Bidirectional
               name: plugins-dir
+            - mountPath: /csi-data-dir
+              name: csi-data-dir
```
xing-yang: This change should be added to the deployment file in deploy/master/hostpath directory too.

Also in this file, it uses image: quay.io/k8scsi/hostpathplugin:v1.0.1.  That's not going to work for raw block.

aayushrangwala: Right!, added  mountPath in deploy/master/hostpath/csi-hostpath-plugin.yaml and changed the image version to Canary

deploy/kubernetes-1.13/hostpath/csi-hostpath-plugin.yaml
```diff
@@ -37,7 +37,7 @@ spec:
       hostNetwork: true
       containers:
         - name: node-driver-registrar
-          image: quay.io/k8scsi/csi-node-driver-registrar:v1.0.2
+          image: quay.io/k8scsi/csi-node-driver-registrar:canary
```
pohly: :-1: 

This is the deployment which defines what sidecars are meant to run on Kubernetes 1.13. For all of the sidecars that means that we have to stick with some v1.0.x version, because only those use the CRDs for CSINodeInfo and CSIDriverInfo.

deploy/kubernetes-1.13/hostpath/csi-hostpath-plugin.yaml
```diff
@@ -63,7 +63,7 @@ spec:
             name: csi-data-dir
 
         - name: hostpath
-          image: quay.io/k8scsi/hostpathplugin:v1.1.0-rc1
+          image: quay.io/k8scsi/hostpathplugin:canary
```
pohly: This however is fine, because the latest driver works on 1.13.

We just need to remember to go back to a fixed version again before the next driver release.

pkg/hostpath/controllerserver.go
```diff
@@ -347,7 +347,14 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 	creationTime := ptypes.TimestampNow()
 	volPath := hostPathVolume.VolPath
 	file := snapshotRoot + snapshotID + ".tgz"
```
pohly: A better fix is to use `path.Join` instead of appending the slash to the directory constants.

deploy/kubernetes-1.13/hostpath/csi-hostpath-plugin.yaml
aayushrangwala: Yes Correct, till then we must be having a hostpathplugin image fixed version which will contain the current code changes

deploy/kubernetes-1.13/hostpath/csi-hostpath-plugin.yaml
aayushrangwala: Understood, The external attacher uses CSINode API and CSINodeInfo handeling is removed in release-1.1

pkg/hostpath/controllerserver.go
aayushrangwala: Are you suggesting like this:
`filePath := []string{snapshotRoot, "/", snapshotID, ".tgz"}
	file := strings.Join(filePath, "")`
If yes, then we will need to change in DeleteSnapshot() also

xing-yang:
@pohly do you know why csi-restore.yaml and csi-snapshot.yaml are under https://github.com/kubernetes-csi/csi-driver-host-path/tree/master/deploy/snapshot, not under https://github.com/kubernetes-csi/csi-driver-host-path/tree/master/examples?  It doesn't look like they are used by the deployment script.  Even if they are needed under deployment, I think we should add a link under examples to point to them.

I see that csi-hostpath-snapshotclass.yaml is under github.com/kubernetes-csi/csi-driver-host-path/deploy/master/snapshotter and it is used by the deployment script, so it should stay there.  We should still add a link under examples to point to it.

pohly:
> do you know why csi-restore.yaml and csi-snapshot.yaml are under https://github.com/kubernetes-csi/csi-driver-host-path/tree/master/deploy/snapshot, not under https://github.com/kubernetes-csi/csi-driver-host-path/tree/master/examples?

I'm not sure exactly. Perhaps the idea was that the content might be specific to the actual deployment. If that isn't the case right now, then we should move them under `examples`. If they ever become different for different deployments, we can still create different versions.

pkg/hostpath/controllerserver.go
aayushrangwala: commited

deploy/kubernetes-1.13/hostpath/csi-hostpath-plugin.yaml
aayushrangwala: commited

aayushrangwala:
@xing-yang  and @pohly  committed with the changes suggested by you, Please have a look

xing-yang:
@msau42 PTAL

PR #43, opened 2019-04-15
deploy: intermediate update to v1.1.0-rc1

pohly:
**What type of PR is this?**
/kind feature

With hostpath driver 1.1.0, we will replace the older v1.0.1 release
for all deployments. This is something that we can start testing in
the CI periodic jobs by updating to the v1.1.0-rc1 image.

This makes the deployment on the master branch temporarily use a pre-release image. That is okay according to our policy for these deployments, which is "master might be broken, users should use a tagged release".

```release-note
Kubernetes 1.13: use hostpath v1.0.0-rc1
```

### DIFF
--- a/deploy/kubernetes-1.13/hostpath/csi-hostpath-plugin.yaml
+++ b/deploy/kubernetes-1.13/hostpath/csi-hostpath-plugin.yaml
@@ -63,7 +63,7 @@ spec:
             name: csi-data-dir
 
         - name: hostpath
-          image: quay.io/k8scsi/hostpathplugin:v1.0.1
+          image: quay.io/k8scsi/hostpathplugin:v1.1.0-rc1
           args:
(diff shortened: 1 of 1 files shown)


pohly:
@msau42 we discussed in one of the previous meetings and again today that we should replace v1.0.1.

It occurred to me that we should try to test this in advance, which we can do by updating to the -rc1. This is not covered by the `pull-` tests, none of them use the unmodified deployments. We'll have to watch jobs like https://k8s-testgrid.appspot.com/sig-storage-csi#1-13-on-master once this is merged.

pohly:
Periodic testing now is almost clean for `13-on-*` jobs. One fix is missing in the v1.1.0-rc1 image (9c696618c38af14ea7a6615be7d63b2bb2c98c2c), which causes one last test failure.

PR #44, opened 2019-04-19
Update sidecar images to v1.1.0

xing-yang:
**What type of PR is this?**
/kind bug

This PR updates sidecar images to v1.1.0.

Fixes #

```release-note
NONE
```

### DIFF
--- a/deploy/master/hostpath/csi-hostpath-plugin.yaml
+++ b/deploy/master/hostpath/csi-hostpath-plugin.yaml
@@ -37,7 +37,7 @@ spec:
       hostNetwork: true
       containers:
         - name: node-driver-registrar
-          image: quay.io/k8scsi/csi-node-driver-registrar:canary # TODO: replace with released version
+          image: quay.io/k8scsi/csi-node-driver-registrar:v1.1.0
           lifecycle:
             preStop:
               exec:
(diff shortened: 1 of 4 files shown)


xing-yang:
@msau42 PTAL

deploy/master/hostpath/csi-hostpath-plugin.yaml
```diff
@@ -63,7 +63,7 @@ spec:
             name: csi-data-dir
 
         - name: hostpath
-          image: quay.io/k8scsi/hostpathplugin:canary # TODO: replace with released version
+          image: quay.io/k8scsi/hostpathplugin:canary # TODO: replace with v1.1.0 when it is cut
```
pohly: I think @msau42 wanted to release v1.1.0 with a .yaml file that references the corresponding image, i.e. v1.1.0. Has that changed?

Switching to the v1.1.0 image before the v1.1.0 release breaks the master branch until that release is complete, but that was seen as a better solution that releasing v1.1.0 with an unmatched .yaml file.



From a CI perspective, that breakage of master will show up in the periodic jobs. Pull jobs are not affected.

The `deploy/kubernetes-1.13` also needs the v1.1.0 hostpath image.

Another change: `deploy/master` should be renamed to `deploy/kubernetes-1.14`. `deploy/master` can be removed. `deploy/latest` needs to point to `deploy/kubernetes-1.14`.

xing-yang: I think @msau42 said we need to do this in steps, cut a v1.1.0-rc first (there is already a rc1 cut a month ago, so this will be rc2), update yaml, then cut v1.1.0, and then update yaml to 1.1.0.  Need to confirm with her on details.

pohly: FWIW, the previous discussion around this starts here: https://github.com/kubernetes/test-infra/pull/11981#discussion_r271175985

PR #48, opened 2019-05-01
release 1.1.0 deployments

pohly:
**What type of PR is this?**
/kind api-change

This is needed because we want the v1.1.0 release of the driver to have the corresponding deployments that use that driver.

The first commit could already be merged now. Let me know if you want to do that and I'll create a separate PR for it. The second commit should be merged only shortly before the actual v1.1.0 release, because merging it will break the deployments and thus the periodic CI testing until the release image becomes available.

```release-note
deployments for Kubernetes 1.13 and 1.14 use the v1.1.0 driver image
```


deploy/kubernetes-1.14/hostpath/csi-hostpath-plugin.yaml
```diff
@@ -63,7 +63,7 @@ spec:
             name: csi-data-dir
 
         - name: hostpath
-          image: quay.io/k8scsi/hostpathplugin:canary # TODO: replace with v1.1.0 when it is cut
```
msau42: do we still want a deployment for master?

pohly: No. The CI only uses "kubernetes-x.yy". Using "master" as directory and "kubernetes-1.14" as symlink was an interim solution to indicate to users that the deployment hadn't been finalized yet. That's no longer the case.

I also can't think of another reason right why we might want to have a separate deployment for Kubernetes master. If someone wants to try out the canary images, then they can do that with `IMAGE_TAG=canary deploy/kubernetes-latest/deploy-hostpath.sh`.

msau42:
Oh one more thing. Can you help prepare a changelog file similar to the sidecars that summarize the important features and fixes for this release?

pohly:
> Oh one more thing. Can you help prepare a changelog file similar to
> the sidecars that summarize the important features and fixes for this
> release?

I had the same thought and actually tried, but the auto-generated
changelog from the release-notes tools is pretty useless for this repo
(just a single entry because most of the PRs don't have release
notes). I then had to stop due to lack of time.

Looking at the commit log I'm not sure which changes need to be
mentioned.

Given that this is a demo driver, perhaps we can simply release 1.1.0
without a changelog and then become more diligent about it for future
releases?

msau42:
That's fine. What were the main changes we made? Added raw block support, improved deployment example, added snapshot examples

PR #56, opened 2019-06-04
Doc update for CSI inline ephemeral feature

vladimirvivien:
**What type of PR is this?**
/kind documentation

This PR documents the support for the inline ephemeral feature in CSI.

```release-note
NONE
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -341,6 +341,66 @@ Follow the following example to create a volume from a volume snapshot:
 > pvc-77324684-c717-11e8-8911-000c2967769a   1Gi        RWO            Delete           Bound    default/hpvc-restore   csi-hostpath-sc            33s
 > ```
 
+## Inline ephemeral support
+As of version 1.15, the CSI Hostpath driver now includes support for inline ephemeral volume.  This means that a volume can be specified directly
+inside a pod spec without the need to use a persistent volume object.  Find more detail about the design of this feature 
+[here](https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/20190122-csi-inline-volumes.md).  
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -341,6 +341,66 @@ Follow the following example to create a volume from a volume snapshot:
 > pvc-77324684-c717-11e8-8911-000c2967769a   1Gi        RWO            Delete           Bound    default/hpvc-restore   csi-hostpath-sc            33s
 > ```
 
+## Inline ephemeral support
+As of version 1.15, the CSI Hostpath driver now includes support for inline ephemeral volume.  This means that a volume can be specified directly
```
msau42: Kubernetes version 1.15 and CSI hostpath driver version X?

README.md
```diff
@@ -341,6 +341,66 @@ Follow the following example to create a volume from a volume snapshot:
...
+inside a pod spec without the need to use a persistent volume object.  Find more detail about the design of this feature 
+[here](https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/20190122-csi-inline-volumes.md).  
+
+You will need to enable the feature gate `CSIInlineVolume` when starting your Kubernets binary components:
```
msau42: Point to kubernetes-csi docs page for instructions on how to enable the feature and also more details about the feature

msau42: I think this feature gate instructions can be handled in the csi-docs page (because it's going to get outdated when it goes to beta/ga)

msau42:
The README is getting pretty long.  We may eventually want to consider breaking up this README into a user guide directory with examples for different features, like snapshots and inline volumes.  Doesn't have to be in this PR though

vladimirvivien:
@msau42 PTAL

PR #58, opened 2019-06-11
Implement Cloning for csi-hostpath driver

### LINKED ISSUE #57: Clone support
With the addition of clone support in kubernetes and in progress for the external-provisioner side car we should support clone operations in the hostpath driver.

j-griffith:
This PR adds support for cloning of hostpath volumes.

**What type of PR is this?**
> /kind feature

This PR adds a simple implementation for clone support to the csi-hostpath driver.

Fixes #57 

Kubernetes changes have merged in to current master, target for Alpha in 1.15
Sidecar changes are here: https://github.com/kubernetes-csi/external-provisioner/pull/220

```release-note
Adds support for cloning of hostpath volumes
```

### DIFF
--- a/pkg/hostpath/controllerserver.go
+++ b/pkg/hostpath/controllerserver.go
@@ -65,6 +65,7 @@ func NewControllerServer(ephemeral bool) *controllerServer {
 				csi.ControllerServiceCapability_RPC_CREATE_DELETE_VOLUME,
 				csi.ControllerServiceCapability_RPC_CREATE_DELETE_SNAPSHOT,
 				csi.ControllerServiceCapability_RPC_LIST_SNAPSHOTS,
+				csi.ControllerServiceCapability_RPC_CLONE_VOLUME,
 			}),
 	}
 }
@@ -178,9 +179,11 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 			snapshotId := contentSource.GetSnapshot().GetSnapshotId()
 			snapshot, ok := hostPathVolumeSnapshots[snapshotId]
 			if !ok {
+				deleteHostpathVolume(volumeID)
 				return nil, status.Errorf(codes.NotFound, "cannot find snapshot %v", snapshotId)
 			}
 			if snapshot.ReadyToUse != true {
+				deleteHostpathVolume(volumeID)
 				return nil, status.Errorf(codes.Internal, "Snapshot %v is not yet ready to use.", snapshotId)
 			}
 			snapshotPath := snapshot.Path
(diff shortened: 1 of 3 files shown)


pkg/hostpath/controllerserver.go
```diff
@@ -191,6 +192,20 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
...
 			}
 		}
+		if srcVolume := contentSource.GetVolume(); srcVolume != nil {
+			srcVolumeID := srcVolume.GetVolumeId()
+			hostPathVolume, ok := hostPathVolumes[srcVolumeID]
+			if !ok {
+				return nil, status.Error(codes.Internal, "source volumeID does not exist, are source/destination in the same storage class?")
+			}
+			srcPath := hostPathVolume.VolPath
+			args := []string{"-r", srcPath + "/*", path + "/"}
```
jsafrane: Would be `-a` more appropriate for 1:1 clone?

pkg/hostpath/controllerserver.go
```diff
@@ -191,6 +192,20 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
...
+			executor := utilexec.New()
+			out, err := executor.Command("cp", args...).CombinedOutput()
+			if err != nil {
+				return nil, status.Error(codes.Internal, fmt.Sprintf("failed pre-populate data (clone) for volume: %v: %s", err, out))
```
jsafrane: Should the driver at least try to delete half-cloned volume on error? Fixing it in the snapshot restore case above would be nice too :-).

pkg/hostpath/controllerserver.go
j-griffith: Indeed!  The one other thing I wanted to check on is handling block mode.  Thanks for pointing this out.

pkg/hostpath/controllerserver.go
j-griffith: For sure.. I'll update both.

jsafrane:
/lgtm
/approve

why there is hold?

jsafrane:
@j-griffith, it seems that csi-sanity fails for cloning.

jsafrane:
/lgtm cancel
until tests are green

j-griffith:
> @j-griffith, it seems that csi-sanity fails for cloning.

There are probably some updates needed there, but we also seem to only have 1.13/14 enabled there; since the code in kubernetes just merged and will be 1.15 I'm not sure (yet) how to pull that in.

pohly:
> There are probably some updates needed there, but we also seem to only have 1.13/14 enabled there; since the code in kubernetes just merged and will be 1.15 I'm not sure (yet) how to pull that in.

csi-sanity testing is independent of the Kubernetes version. It merely runs against the hostpath driver that got deployed into a Kubernetes cluster.

What we have here are simply some failing tests: "can't stat '/csi-data-dir/1a9fcb59-8d35-11e9-aa7d-024285e8710b/*': No such file or directory" and an unexpected error code.

It might very well be that the snapshot test is broken. I don't know whether it ever has been run against a driver that implements snapshotting. The error code mismatch looks like something that can be fixed in the hostpath implementation.

To reproduce the failure:
- create a *clean* `$GOPATH` because .prow.sh will mess with it (it's meant to be run in CI)
- check out your branch in $GOPATH/github.com/kubernetes-csi/csi-driver-host-path
- `CSI_PROW_TESTS=sanity ./.prow.sh`

Once that has failed, you still have a running cluster in KinD and you can debug manually (re-run the test, update the driver, etc.)

To clean up, `kind delete cluster --name=csi-prow` (if you don't have it, a binary was built by the script) and delete that `$GOPATH`.

j-griffith:
> > There are probably some updates needed there, but we also seem to only have 1.13/14 enabled there; since the code in kubernetes just merged and will be 1.15 I'm not sure (yet) how to pull that in.
> 
> csi-sanity testing is independent of the Kubernetes version. It merely runs against the hostpath driver that got deployed into a Kubernetes cluster.
> 
> What we have here are simply some failing tests: "can't stat '/csi-data-dir/1a9fcb59-8d35-11e9-aa7d-024285e8710b/*': No such file or directory" and an unexpected error code.
> 
> It might very well be that the snapshot test is broken. I don't know whether it ever has been run against a driver that implements snapshotting. The error code mismatch looks like something that can be fixed in the hostpath implementation.
> 
> To reproduce the failure:
> 
> * create a _clean_ `$GOPATH` because .prow.sh will mess with it (it's meant to be run in CI)
> * check out your branch in $GOPATH/github.com/kubernetes-csi/csi-driver-host-path
> * `CSI_PROW_TESTS=sanity ./.prow.sh`
> 
> Once that has failed, you still have a running cluster in KinD and you can debug manually (re-run the test, update the driver, etc.)
> 
> To clean up, `kind delete cluster --name=csi-prow` (if you don't have it, a binary was built by the script) and delete that `$GOPATH`.

Thanks @pohly I'm looking into it

j-griffith:
@pohly thanks for the hints, the problem is that in the test case the src directory is empty; so the `cp -a/*` call fails.  now that I know what's going on I can get it fixed up, thanks again for the pointers.

j-griffith:
Missed one because I was using focused test.

j-griffith:
@jsafrane @pohly Should be good to go now, thanks for the help.

PR #39, opened 2019-04-15
update to Go 1.12.4

pohly:

PR #64, opened 2019-07-08
examples + drivername

### LINKED ISSUE #38: Add domain to driver name
Maybe hostpath.kubernetes-csi.io?

pohly:
**What type of PR is this?**
/kind documentation

We want the name to follow the recommendation from the spec and use a domain. While testing this, some issues with the examples were noticed and fixed.

Fixes #38

```release-note
the deployment uses `hostpath.csi.k8s.io` as driver name
```


pohly:
Looks like I need to update csi-release-tools first with a more flexible way of creating the external driver test config. Hard-coding `csi-hostpath` as name no longer works.

/hold

pohly:
@msau42: we previously postponed updating Go to 1.12 (PR #39). Because this PR here updates csi-release-tools, it also includes that change. I lost track of how long we wanted to postpone, though. Probably it was something that should have been in csi-driver-host-path for Kubernetes 1.15?

Anyway, this PR is becoming more complicated than expected. If you want to include just the examples fix, then merge PR #65 instead.

msau42:
Yes we should update go version now

README.md
```diff
@@ -82,7 +82,7 @@ csi-hostpathplugin-0         2/2     Running   0          5m45s
 From the root directory, deploy the application pods including a storage class, a PVC, and a pod which mounts a volume using the Hostpath driver found in directory `./examples`:
 
 ```shell
-$ kubectl create -f ./examples
+$ for i in ./examples/csi-storageclass.yaml ./examples/csi-pvc.yaml ./examples/csi-app.yaml; do kubectl apply -f $i; done
```
msau42: should we explicitly call out each file as a separate step?

msau42:
Is there a way to get rid of this merge commit: 6128d3d61e6f9d598e164bef5687e6df637ee716?

pohly:
> Is there a way to get rid of this merge commit: 6128d3d?

That's the `release-tools` update, which is needed for changing the driver name, and importing with `git subtree` has to be done with merge commits that splices together the external repo and the one it is getting merged into. So, no, this cannot be avoided. I can change the commit message, if that helps.

README.md
pohly: We could. It would be more work to copy-and-paste the whole sequence. I don't have a strong opinion here.

PR #66, opened 2019-07-10
WIP: Add 1.15 testing

msau42:
**What type of PR is this?**
/kind feature

Updates prow.sh to use 1.15 hostpath deployment. As a result, the kind version also has to be upgraded to 0.4.0, which requires 2 major changes to our prow job:
1. We need to use the new `kind.sigs.k8s.io/v1alpha3 Cluster' API object, which uses `kubeadm.k8s.io/v1beta2` which only works for 1.15+.  As a result, our alpha jobs can only run on 1.15+
2. kind 0.4.0 supports very specific patch images of 1.15, 1.14, 1.13, etc, which are not the ones we have configured in our prow jobs.  As a result, I have added logic in the prow job to override the patch versions.

Fixes #

```release-note
NONE
```

### DIFF
--- a/release-tools/prow.sh
+++ b/release-tools/prow.sh
@@ -183,7 +183,7 @@ configvar CSI_PROW_WORK "$(mkdir -p "$GOPATH/pkg" && mktemp -d "$GOPATH/pkg/csip
 #
 # When no deploy script is found (nothing in `deploy` directory,
 # CSI_PROW_HOSTPATH_REPO=none), nothing gets deployed.
-configvar CSI_PROW_HOSTPATH_VERSION fc52d13ba07922c80555a24616a5b16480350c3f "hostpath driver" # pre-1.1.0
+configvar CSI_PROW_HOSTPATH_VERSION "v1.1.0" "hostpath driver"
 configvar CSI_PROW_HOSTPATH_REPO https://github.com/kubernetes-csi/csi-driver-host-path "hostpath repo"
 configvar CSI_PROW_DEPLOYMENT "" "deployment"
 configvar CSI_PROW_HOSTPATH_DRIVER_NAME "csi-hostpath" "the driver (aka provisioner) name of the chosen hostpath driver"
--- a/release-tools/prow.sh
+++ b/release-tools/prow.sh
@@ -470,6 +470,14 @@ kind: Cluster
 apiVersion: kind.sigs.k8s.io/v1alpha3
 nodes:
 - role: control-plane
+EOF
+
+    # kubeadm has API dependencies between apiVersion and Kubernetes version
+    # 1.15+ requires kubeadm.k8s.io/v1beta2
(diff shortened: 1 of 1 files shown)


pohly:
There's something odd in the job config: pull-kubernetes-csi-csi-driver-host-path-unit ran with `CSI_PROW_TESTS=parallel`. It should have execute the unit tests instead. Hmm, this looks right: https://github.com/kubernetes/test-infra/blob/0e3897d9c59bb364c306f671f914c83fc4e401dc/config/jobs/kubernetes-csi/csi-driver-host-path/csi-driver-host-path-config.yaml#L221-L251

Ah, you are overriding that in `.prow.sh`. That explains it :smile: 

> Kubernetes versions may also have to be updated because kind 0.4.0 only supports specific versions.

Does it still support the ones we care about (1.13, 1.14 and now 1.15)? It shouldn't matter too much which version exactly that is.

> Unsure if there's going to be an atomic way to update the version without breaking pull and ci jobs temporarily

It should be atomic: once this PR gets merged, the existing pull and periodic CI jobs will use the config from the updated master branch. We just need to be sure that the updated config also works for them. Once you remove the `.prow.sh` overrides, we'll test that for the mandatory pull jobs when they run.

What isn't atomic is adding the new 1.15 jobs. One way of doing that is:
- define the new jobs in test-infra without making them mandatory
- run the new `pull-kubernetes-csi-csi-driver-host-path-1-15-on-kubernetes-1-15` manually in this PR via the `test` command
- merge this PR and similar ones for the other sidecars
- update the new jobs to make them mandatory

msau42:
@pohly kind 0.4.0 only supports 1.15.0, 1.14.3, and 1.13.7, which is different from the versions we configured in our test-infra.

So I think in order to make this switch, we will have to temporarily disable/optional the pull jobs and change the versions in the jobs, so that we can merge this PR.  Do you see any better ways to do this?

msau42:
> Patrick Ohly kind 0.4.0 only supports 1.15.0, 1.14.3, and 1.13.7, which is different from the versions we configured in our test-infra.
> 
> So I think in order to make this switch, we will have to temporarily disable/optional the pull jobs and change the versions in the jobs, so that we can merge this PR. Do you see any better ways to do this?

My last commit attempts to workaround this by overriding the patch versions no matter what.  LMK if you think that's acceptable.

pohly:
> My last commit attempts to workaround this by overriding the patch versions no matter what. LMK if you think that's acceptable.

Yes, that's okay. I'm just wondering whether we should change the semantic of `CSI_PROW_KUBERNETES_VERSION` and make it's value just `<major>.<minor>`. OTOH, for future, unknown versions that aren't getting patched we still need `<major>.<minor>.<patch>`. Let's just keep it as-is.

msau42:
/close
Addressed by #70

PR #67, opened 2019-07-10
Inline and persistent volumes

pohly:
**What type of PR is this?**
/kind feature

Supporting normal and inline volumes in the same driver deployment makes it possible to test both as part of the existing Prow jobs without having to change the driver deployment.

Related-to: https://github.com/kubernetes/kubernetes/issues/79624

```release-note
The -ephemeral parameter (currently alpha) is still supported, but only needed for Kubernetes 1.15 and will be removed once Kubernetes 1.15 stops being supported. On Kubernetes 1.16, the same deployment supports normal persistent volumes and inline ephemeral volumes.
```

### DIFF
--- a/cmd/hostpathplugin/main.go
+++ b/cmd/hostpathplugin/main.go
@@ -48,6 +48,10 @@ func main() {
 		return
 	}
 
+	if *ephemeral {
+		fmt.Fprintln(os.Stderr, "Deprecation warning: The ephemeral flag is deprecated and should only be used when deploying on Kubernetes 1.15. It will be removed in the future.")
+	}
+
 	handle()
 	os.Exit(0)
 }
--- a/cmd/hostpathplugin/main.go
+++ b/cmd/hostpathplugin/main.go
@@ -33,7 +33,7 @@ var (
 	endpoint    = flag.String("endpoint", "unix://tmp/csi.sock", "CSI endpoint")
 	driverName  = flag.String("drivername", "hostpath.csi.k8s.io", "name of the driver")
 	nodeID      = flag.String("nodeid", "", "node id")
-	ephemeral   = flag.Bool("ephemeral", false, "deploy in ephemeral mode")
+	ephemeral   = flag.Bool("ephemeral", false, "publish volumes in ephemeral mode even if kubelet did not ask for it (only needed for Kubernetes 1.15)")
 	showVersion = flag.Bool("version", false, "Show version.")
 	// Set by the build process
 	version = ""
--- a/README.md
+++ b/README.md
@@ -346,7 +346,7 @@ Follow the following example to create a volume from a volume snapshot:
 As of version 1.15 of Kubernetes, the CSI Hostpath driver (starting with version 1.0.1) now includes support for inline ephemeral volume.  This means that a volume can be specified directly inside a pod spec without the need to use a persistent volume object.
 Find out how to enable or create a CSI inline driver [here](https://kubernetes-csi.github.io/docs/ephemeral-local-volumes.html)
 
-To test this feature, redeploy the CSI Hostpath plugin YAML by updating the `hostpath` container to use  the inline ephemeral mode by setting the `ephemeral` flag, of the driver binary, to true as shown in the following setup:
+To test this feature on Kubernetes 1.15, redeploy the CSI Hostpath plugin YAML by updating the `hostpath` container to use  the inline ephemeral mode by setting the `ephemeral` flag, of the driver binary, to true as shown in the following setup:
 
 ```yaml
 kind: DaemonSet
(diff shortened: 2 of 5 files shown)


pohly:
/hold

- this uses the updated driver name, so PR #64 should be merged first
- this is targeting Kubernetes 1.16: once merged, the driver will no longer work with inline volumes in Kubernetes  1.15

cmd/hostpathplugin/main.go
```diff
@@ -33,7 +34,7 @@ var (
 	endpoint    = flag.String("endpoint", "unix://tmp/csi.sock", "CSI endpoint")
 	driverName  = flag.String("drivername", "hostpath.csi.k8s.io", "name of the driver")
 	nodeID      = flag.String("nodeid", "", "node id")
-	ephemeral   = flag.Bool("ephemeral", false, "deploy in ephemeral mode")
+	_           = deprecatedflags.Add("ephemeral")
```
kfox1111: This PR looks good to me but maybe this.

Its marked deprecated , but actually broken, as I don't see it ever honoured anywhere. This is likely not a big issue, as its basically just a testing driver anyway. So maybe just delete the flag entirely?

I just now dug through what the deprecatedflags does, and it is ignored. This is ok I guess, but the way deprecated is being defined here is not how the rest of kubernetes usually uses it? It confused me, so likely to confused others. I believe deprecated elsewhere means 'it still works but you shouldn't use it. Its likely to go away entirely at some point soon", not "Its already broken, stop using it". Having it actually break due to the missing flag may be safer/more consistent for the user.

pohly: We've deprecated flags the same way elsewhere (for example, in https://github.com/kubernetes-csi/external-provisioner/blob/ef483860f8533dd125b1a059733c6a33bea106c6/cmd/csi-provisioner/csi-provisioner.go#L53-L61). I don't have a strong opinion on whether it is better to print an explicit warning for unused flags or fail when they are used.

@jsafrane, any comment?

jsafrane: I agree with deprecation. People may have yaml files where the flag is set and it would stop working. Now they have 3 months (or even longer) to update them.

kfox1111: If they launch the new container in a 1.15 cluster, it will look to work, but be broken. 1.15 is not deprecated, its unsupported.

If it actually died, then they would know why the driver is broken much quicker. logs go out to container logs where they might not notice the warning for a while. While dieing actually shows up right away as, its broken.

Again though, for this particular driver, probably doesn't matter. Just thinking in general about the process.

pohly: @kfox1111 has a point here. I'm not too worried about people deploying this incorrectly, but having to maintain two versions of the driver or removing support for testing this on 1.15 both are not appealing.

It's much easier to keep the `--ephemeral` parameter: if set, always act in ephemeral mode, without relying on the hint from kubelet (works with ephemeral volumes and 1.15). If not set, use the hint to detect ephemeral volumes (works with ephemeral volumes and 1.16) and if not set of false, assume normal persistent volumes (works with 1.15 and 1.16).

pohly:
@kfox1111, @jsafrane: I force-pushed a much simpler commit which still has the `--ephemeral` parameter. Because I didn't import `deprecatedflags`, none of the dependencies had to be changed. We should still update, but that can be done separately.

jsafrane:
/approve
@kfox1111, @msau42, WDYT?

pohly:
I had forgotten to update the changelog. I just did that now.

pohly:
/hold cancel

The driver name change was merged and the driver will still work on 1.15.

cmd/hostpathplugin/main.go
```diff
@@ -33,7 +33,7 @@ var (
 	endpoint    = flag.String("endpoint", "unix://tmp/csi.sock", "CSI endpoint")
 	driverName  = flag.String("drivername", "hostpath.csi.k8s.io", "name of the driver")
 	nodeID      = flag.String("nodeid", "", "node id")
-	ephemeral   = flag.Bool("ephemeral", false, "deploy in ephemeral mode")
+	ephemeral   = flag.Bool("ephemeral", false, "publish volumes in ephemeral mode even if kubelet did not ask for it (only needed for Kubernetes 1.15)")
```
kfox1111: maybe add at line 44:
if *epemeral {
  fmt.Fprintf(os.Stderr, "Deprecation warning: The ephemeral flag is deprecated and will be removed in the next major version.")
}

kfox1111:
This PR looks great. Only one little change if we want to remove the flag some day.

cmd/hostpathplugin/main.go
pohly: I added this warning: `Deprecation warning: The ephemeral flag is deprecated and should only be used when deploying on Kubernetes 1.15. It will be removed in the future.`

Wording is a bit different - I wanted to make sure that people don't get confused and remove it also while working with 1.15 and I wanted to avoid unnecessary details about when that'll be.

kfox1111:
Perfect. Thanks. :)

pohly:
@jsafrane: can you lgtm based on @kfox1111's review?

PR #78, opened 2019-07-25
Remove external-attacher and add CSIDriver

jsafrane:
/kind cleanup

Initially, I modified also 1.13-1.15, but it looks like it causes test failures.

```release-note
Removed external-attacher usage
```

### DIFF
--- a/deploy/kubernetes-1.15-no-attacher/README.md
+++ b/deploy/kubernetes-1.15-no-attacher/README.md
@@ -1,0 +1,9 @@
+The deployment for Kubernetes 1.15 uses CSI 1.0 and thus is
+incompatible with Kubernetes < 1.13.
+
+The sidecars depend on 1.15 API changes for migration and resizing,
+and 1.14 API changes for CSIDriver and CSINode.
+
+In contrast to the normal deployment, this deployment does not use the
+external-attacher sidecar. Instead, it creates a CSIDriver instance
+that tells Kubernetes not to use attach (ControllerPublishVolume) at all.
--- a/deploy/kubernetes-1.15-no-attacher/deploy-hostpath.sh
+++ b/deploy/kubernetes-1.15-no-attacher/deploy-hostpath.sh
@@ -1,0 +1,1 @@
+../util/deploy-hostpath-no-attacher.sh
--- a/deploy/kubernetes-1.15-no-attacher/hostpath/csi-hostpath-csidriver.yaml
+++ b/deploy/kubernetes-1.15-no-attacher/hostpath/csi-hostpath-csidriver.yaml
@@ -1,0 +1,7 @@
+apiVersion: storage.k8s.io/v1beta1
+kind: CSIDriver
+metadata:
+  name: hostpath.csi.k8s.io
+spec:
+  attachRequired: false
+  podInfoOnMount: false
--- a/deploy/kubernetes-1.15-no-attacher/hostpath/csi-hostpath-plugin.yaml
+++ b/deploy/kubernetes-1.15-no-attacher/hostpath/csi-hostpath-plugin.yaml
@@ -1,0 +1,138 @@
+# Service defined here, plus serviceName below in StatefulSet,
+# are needed only because of condition explained in
+# https://github.com/kubernetes/kubernetes/issues/69608
+
+kind: Service
+apiVersion: v1
+metadata:
+  name: csi-hostpathplugin
+  labels:
+    app: csi-hostpathplugin
+spec:
+  selector:
+    app: csi-hostpathplugin
+  ports:
+    - name: dummy
+      port: 12345
+---
+kind: StatefulSet
+apiVersion: apps/v1
+metadata:
+  name: csi-hostpathplugin
+spec:
+  serviceName: "csi-hostpathplugin"
+  # One replica only:
+  # Host path driver only works when everything runs
+  # on a single node. We achieve that by starting it once and then
+  # co-locate all other pods via inter-pod affinity
+  replicas: 1
+  selector:
+    matchLabels:
+      app: csi-hostpathplugin
+  template:
+    metadata:
+      labels:
+        app: csi-hostpathplugin
+    spec:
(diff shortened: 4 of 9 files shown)


deploy/kubernetes-1.16/README.md
```diff
@@ -0,0 +1,7 @@
+The deployment for Kubernetes 1.16 uses CSI 1.0 and thus is
+incompatible with Kubernetes < 1.13.
+
+The sidecars depend on 1.15 API changes for migration and resizing,
+and 1.14 API changes for CSIDriver and CSINode.
+However the hostpath driver doesn't use those features, so this
+deployment can work on older Kubernetes versions.
```
pohly: We need to explain here why we have a separate deployment for 1.16.
Has something changed in the sidecars?

One reason that I have for wanting a separate deployment is adding `CSIDriver.Spec.Mode` "persistent+ephemeral". That field does not exist in 1.15 yet and for 1.16 would still need to pass review (https://github.com/kubernetes/kubernetes/pull/80568).

Independently of you I just created a deployment for 1.16 and in it's README wrote:
```
The deployment for Kubernetes 1.16 contains a CSIDriver object which
uses "mode: persistent+ephemeral", which is a new field in 1.16, and
therefore this deployment does not work on older Kubernetes releases.
```

I wasn't sure whether using the "mode" field with 1.15 would cause problems, so I tried it - it does (validation error).

deploy/kubernetes-1.16/hostpath/csi-hostpath-csidriver.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: storage.k8s.io/v1beta1
+kind: CSIDriver
+metadata:
+  name: hostpath.csi.k8s.io
+spec:
+  attachRequired: false
+  podInfoOnMount: false
```
pohly: Once it is sure that the new field lands in 1.16, I'd like to have `mode: persistent+ephemeral` here.

msau42:
If we remove external-attacher, does that mean we will lose test coverage on external-attacher repo?

deploy/kubernetes-1.16/README.md
jsafrane: I had no special intention for creating deployment for 1.16 other than removing the attacher + adding CSIDriver.

I think the text above reflects *current* state of the driver and you can add `mode: persistent+ephemeral` when it gets merged in Kubernetes.

deploy/kubernetes-1.16/hostpath/csi-hostpath-csidriver.yaml
jsafrane: Sure, no problem.

pohly:
> If we remove external-attacher, does that mean we will lose test coverage on external-attacher repo?

Yes. For full coverage we would need two different deployments (one with external-attacher, one without) and additional Prow job(s) that actually use the non-default deployment.

jsafrane:
Hmm, the external-attacher uses a very different code path when its CSI driver is not attachable. Moving to mock would be perhaps more useful.

pohly:
> Hmm, the external-attacher uses a very different code path when its CSI driver is not attachable. Moving to mock would be perhaps more useful.

The mock driver wouldn't be usable for some of the "real" storage tests, like writing and reading data.

Perhaps the hostpath driver should be made attachable, even if it is just with some stub operation?

pohly:
Jan Šafránek <notifications@github.com> writes:
> I had no special intention for creating deployment for 1.16 other than
> removing the attacher + adding CSIDriver.

If that is the only difference, then it would be better to create a
deploy/kubernetes-1.15-no-attacher where the only diff against
deploy/kubernetes-1.15 is that change.

Then kubernetes-1.16 can be added once it is needed.

> I think the text above reflects *current* state of the driver and you
> can add `mode: persistent+ephemeral` when it gets merged in
> Kubernetes.

Yes, I just wanted to mention that this will be one reason why we'll
need a separate 1.16 deployment - eventually...

jsafrane:
Moved into deploy/kubernetes-1.15-no-attacher.

Purpose of this PR is to have canonical yaml files for kubernetes/test/e2e, I am not sure that pretending it's attachable is good way to go. I'd like to have tests with a real attachable driver (=GCE) and non-attachable one (=hostpath).

deploy/kubernetes-1.15-no-attacher/README.md
```diff
@@ -0,0 +1,9 @@
+The deployment for Kubernetes 1.15 uses CSI 1.0 and thus is
+incompatible with Kubernetes < 1.13.
+
+The sidecars depend on 1.15 API changes for migration and resizing,
+and 1.14 API changes for CSIDriver and CSINode.
+
+This deployment does not use external-attacher sidecar, it creates
```
pohly: Language tweak: "In contrast to the normal deployment, this deployment does not use the external-attacher sidecar. Instead, it creates a CSIDriver instance..."

jsafrane: Fixed

msau42:
I would like to prioritize testing of real attach path in the attacher sidecar pull job.  Most of the changes we are making to the attacher now are impacting attachable drivers so I think that is more important to test than the non attachable case.  

We can still test non attachable case in CI

pohly:
> I would like to prioritize testing of real attach path in the attacher
> sidecar pull job.

But against which driver?

Currently the only driver deployment that we have which works in KinD is
for the csi-driver-host-path.

msau42:
Correct, so I'm suggesting we add some mode to hostpath driver that can be toggled.

pohly:
> Correct, so I'm suggesting we add some mode to hostpath driver that can be toggled.

Yes, I agree and proposed the same earlier (https://github.com/kubernetes-csi/csi-driver-host-path/pull/78#issuecomment-515362002).

I've filed https://github.com/kubernetes-csi/csi-driver-host-path/issues/80

PR #82, opened 2019-07-31
Fix cp arg expansion bug in cloning implementation

### LINKED ISSUE #81: Using `*` in cp commands is a bad idea
Passing `cp -a /csi-data-dir/xxxx/*` on some shells will actually attempt to try and find a file named `*` to copy.  We should instead be using the cp specific syntax and using `/.` instead.

j-griffith:
**What type of PR is this?**

bug

The clone implementation is just using `cp /*` can result in trying to actually find/copy a file named `*`.  The result in some container images is a stat error for no such file or directory.  Instead we should be using the cp notation of `/.` so we're portable and always work.

Fixes #81

```release-note
NONE
```

### DIFF
--- a/pkg/hostpath/hostpath.go
+++ b/pkg/hostpath/hostpath.go
@@ -293,7 +293,7 @@ func loadFromVolume(srcVolumeId, destPath string) error {
 
 	// If the source hostpath volume is empty it's a noop and we just move along, otherwise the cp call will fail with a a file stat error DNE
 	if !isEmpty {
-		args := []string{"-a", srcPath + "/*", destPath + "/"}
+		args := []string{"-a", srcPath + "/.", destPath + "/"}
 		executor := utilexec.New()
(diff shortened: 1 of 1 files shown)


pkg/hostpath/hostpath.go
```diff
@@ -293,7 +293,7 @@ func loadFromVolume(srcVolumeId, destPath string) error {
 
 	// If the source hostpath volume is empty it's a noop and we just move along, otherwise the cp call will fail with a a file stat error DNE
 	if !isEmpty {
-		args := []string{"-a", srcPath + "/*", destPath + "/"}
+		args := []string{"-a", srcPath + "/.", destPath + "/"}
```
msau42: do you need -R?

j-griffith: Nah, `-a` is the equivalent of `-dR`

PR #85, opened 2019-08-14
implement optional ControllerPublishVolume

### LINKED ISSUE #80: implement optional ControllerPublishVolume
The csi-driver-host-path deployment with external-attacher is used to test that external-attacher works. However, because currently csi-driver-host-path doesn't actually implement `ControllerPublishVolume`, external-attacher uses a simplified code path (https://github.com/kubernetes-csi/csi-driver-host-path/pull/78#issuecomment-515359785).

To test the actual code path for drivers that support attach, csi-driver-host-path should get a boolean command line flag `-controllerPublishVolume` that, if set, enables an implementation of `ControllerPublishVolume`. That implementation doesn't need to do anything besides ensuring that `ControllerPublishVolume` and `ControllerUnpublishVolume` are called correctly. The node operations must check that `ControllerPublishVolume` was called. This can be done by creating some `publish_context` that is getting checked.

zhucan:
**What type of PR is this?**
 /kind cleanup

To test ControllerPublishVolume and ControllerUnpublishVolume could be called
Fixes #80 

@pohly 

```release-note
implement optional ControllerPublishVolume
```

### DIFF
--- a/cmd/hostpathplugin/main.go
+++ b/cmd/hostpathplugin/main.go
@@ -36,7 +36,8 @@ var (
 	ephemeral   = flag.Bool("ephemeral", false, "publish volumes in ephemeral mode even if kubelet did not ask for it (only needed for Kubernetes 1.15)")
 	showVersion = flag.Bool("version", false, "Show version.")
 	// Set by the build process
-	version = ""
+	version    = ""
+	controller = flag.Bool("controller", false, "Open ControllerPublishVolume and ControllerUnpublishVolume")
 )
 
 func main() {
@@ -57,7 +58,7 @@ func main() {
 }
 
 func handle() {
-	driver, err := hostpath.NewHostPathDriver(*driverName, *nodeID, *endpoint, version, *ephemeral)
+	driver, err := hostpath.NewHostPathDriver(*driverName, *nodeID, *endpoint, version, *ephemeral, *controller)
 	if err != nil {
 		fmt.Printf("Failed to initialize driver: %s", err.Error())
 		os.Exit(1)
--- a/pkg/hostpath/controllerserver.go
+++ b/pkg/hostpath/controllerserver.go
@@ -52,9 +52,22 @@ type controllerServer struct {
 	caps []*csi.ControllerServiceCapability
 }
 
-func NewControllerServer(ephemeral bool) *controllerServer {
+func NewControllerServer(ephemeral bool, controller bool) *controllerServer {
 	if ephemeral {
 		return &controllerServer{caps: getControllerServiceCapabilities(nil)}
+	}
+
+	if controller {
+		return &controllerServer{
+			caps: getControllerServiceCapabilities(
+				[]csi.ControllerServiceCapability_RPC_Type{
+					csi.ControllerServiceCapability_RPC_CREATE_DELETE_VOLUME,
+					csi.ControllerServiceCapability_RPC_CREATE_DELETE_SNAPSHOT,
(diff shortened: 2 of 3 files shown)


cmd/hostpathplugin/main.go
```diff
@@ -36,7 +36,8 @@ var (
 	ephemeral   = flag.Bool("ephemeral", false, "publish volumes in ephemeral mode even if kubelet did not ask for it (only needed for Kubernetes 1.15)")
 	showVersion = flag.Bool("version", false, "Show version.")
 	// Set by the build process
-	version = ""
+	version    = ""
+	controller = flag.Bool("controller", false, "Open ControllerPublishVolume and ControllerUnpublishVolume")
```
msau42: i would make the name more specific, like "controllerPublish".  and the description "implements PUBLISH_UNPUBLISH_VOLUME ControllerServiceCapability"

pkg/hostpath/controllerserver.go
```diff
@@ -52,10 +52,23 @@ type controllerServer struct {
...
 }
 
-func NewControllerServer(ephemeral bool) *controllerServer {
+func NewControllerServer(ephemeral bool, controller bool) *controllerServer {
 	if ephemeral {
 		return &controllerServer{caps: getControllerServiceCapabilities(nil)}
 	}
+
+	if controller {
+		return &controllerServer{
```
msau42: Instead of duplicating the array, I would append to the default array if the flag is enabled

pkg/hostpath/controllerserver.go
```diff
@@ -234,11 +247,13 @@ func (cs *controllerServer) ValidateVolumeCapabilities(ctx context.Context, req
 }
 
 func (cs *controllerServer) ControllerPublishVolume(ctx context.Context, req *csi.ControllerPublishVolumeRequest) (*csi.ControllerPublishVolumeResponse, error) {
-	return nil, status.Error(codes.Unimplemented, "")
+	glog.V(4).Infof("ControllerPublishVolume %v", req)
```
msau42: In the issue, @pohly suggested that we persist some metadata here so that in NodePublishVolume, we can verify that the ControllerPublishVolume was called.

zhucan: for example? @msau42

cmd/hostpathplugin/main.go
zhucan: copy that, thanks

pkg/hostpath/controllerserver.go
zhucan: copy that, thanks

pkg/hostpath/controllerserver.go
zhucan: Check request? @msau42  @pohly

zhucan: we can know whether ControllerPublishVolume was called through the print of logs

msau42: I think the suggestion patrick had was to actually write some metadata to a file and then during NodePublishVolume, we can programatically verify it.

hoyho: > I think the suggestion patrick had was to actually write some metadata to a file and then during NodePublishVolume, we can programatically verify it.

How about set a/some k-v in ControllerPublishVolumeResponse field PublishContext
e.g.
```
	return &csi.ControllerPublishVolumeResponse{
		PublishContext: map[string]string{
			"ControllerPublished":"true",
		},
	}, nil
```
During NodePublishVolume, we can verify it by checking parameter `NodePublishVolumeRequest.PublishContext`

PR #87, opened 2019-08-19
[WIP] Add volume expansion support

bertinatto:
**What type of PR is this?**

/kind feature

This PR adds volume expansion support to the hostpath CSI driver.

Fixes #

```release-note
Added support for volume expansion.
```


pkg/hostpath/controllerserver.go
```diff
@@ -63,6 +63,7 @@ func NewControllerServer(ephemeral bool) *controllerServer {
 				csi.ControllerServiceCapability_RPC_CREATE_DELETE_SNAPSHOT,
 				csi.ControllerServiceCapability_RPC_LIST_SNAPSHOTS,
 				csi.ControllerServiceCapability_RPC_CLONE_VOLUME,
+				csi.ControllerServiceCapability_RPC_EXPAND_VOLUME,
```
msau42: deployment spec should be updated to add the external-resizer sidecar

bertinatto: We need to release a new version of the driver before updating the manifest, otherwise we'll break the deployment script.

bertinatto: Created #90 for that.

msau42:
/approve
Will let @gnufied do final review

gnufied:
discussed with @bertinatto offline. I think we should also implement `NodeExpandVolume` so as we can do complete test. But I am thinking we can also implement that as a follow up.

pkg/hostpath/nodeserver.go
```diff
@@ -277,10 +277,28 @@ func (ns *nodeServer) NodeGetCapabilities(ctx context.Context, req *csi.NodeGetC
...
+// NodeExpandVolume is only implemented so the driver can be used for e2e testing.
+func (ns *nodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandVolumeRequest) (*csi.NodeExpandVolumeResponse, error) {
+
+	volID := req.GetVolumeId()
+	if len(volID) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "Volume ID not provided")
+	}
+
+	return &csi.NodeExpandVolumeResponse{}, nil
+}
```
gnufied: I think we should also verify if provided `volume_path` was correct. This can be done by - comparing supplied `volume_path` and `target_path` associated with the volume.

pkg/hostpath/controllerserver.go
msau42: actually, how does this break the deployment script?  The pull jobs running in this PR should be testing a custom built image of the hostpath driver that includes your change.  

And before we tag a new image, when the CI jobs run, it will run the older version of the driver that doesn't implement expand volume. Does that break the resizer?

bertinatto: That's correct, the resizer fails to start because the version of driver specified in the manifest doesn't implement expand volume.

PR #88, opened 2019-08-20
enable topology support

### LINKED ISSUE #54: topology support
The hostpath driver should support topology on Kubernetes 1.14. Then applications no longer need to be scheduled manually onto the right node and we can remove that workaround from the E2E testing in Kubernetes.

It also might make the driver complete enough to serve as default storage driver in a KinD cluster.

mucahitkurt:
**What type of PR is this?**
/kind feature

Enable topology support of the driver

Fixes #54 

I have not tested well yet, opened the PR to get feedbacks if I miss anything, I will try to test the change.

```release-note
Enable topology support of CSI hostpath driver.
```
cc @msau42

### DIFF
--- a/pkg/hostpath/identityserver.go
+++ b/pkg/hostpath/identityserver.go
@@ -68,6 +68,13 @@ func (ids *identityServer) GetPluginCapabilities(ctx context.Context, req *csi.G
 					},
 				},
 			},
+			{
+				Type: &csi.PluginCapability_Service_{
+					Service: &csi.PluginCapability_Service{
+						Type: csi.PluginCapability_Service_VOLUME_ACCESSIBILITY_CONSTRAINTS,
+					},
+				},
+			},
 		},
 	}, nil
 }
--- a/pkg/hostpath/nodeserver.go
+++ b/pkg/hostpath/nodeserver.go
@@ -31,6 +31,8 @@ import (
 	"k8s.io/kubernetes/pkg/volume/util/volumepathhandler"
 )
 
+const TopologyKeyNode = "topology.hostpath.csi/node"
+
 type nodeServer struct {
 	nodeID    string
 	ephemeral bool
@@ -261,8 +263,13 @@ func (ns *nodeServer) NodeUnstageVolume(ctx context.Context, req *csi.NodeUnstag
 
 func (ns *nodeServer) NodeGetInfo(ctx context.Context, req *csi.NodeGetInfoRequest) (*csi.NodeGetInfoResponse, error) {
 
+	topology := &csi.Topology{
+		Segments: map[string]string{TopologyKeyNode: ns.nodeID},
+	}
+
 	return &csi.NodeGetInfoResponse{
-		NodeId: ns.nodeID,
+		NodeId:             ns.nodeID,
+		AccessibleTopology: topology,
 	}, nil
 }
 
--- a/deploy/kubernetes-1.14/hostpath/csi-hostpath-provisioner.yaml
+++ b/deploy/kubernetes-1.14/hostpath/csi-hostpath-provisioner.yaml
@@ -45,6 +45,7 @@ spec:
             - -v=5
             - --csi-address=/csi/csi.sock
             - --connection-timeout=15s
+            - --feature-gates=Topology=true
           volumeMounts:
             - mountPath: /csi
               name: socket-dir
--- a/deploy/kubernetes-1.15/hostpath/csi-hostpath-provisioner.yaml
+++ b/deploy/kubernetes-1.15/hostpath/csi-hostpath-provisioner.yaml
@@ -45,6 +45,7 @@ spec:
             - -v=5
             - --csi-address=/csi/csi.sock
             - --connection-timeout=15s
+            - --feature-gates=Topology=true
           volumeMounts:
             - mountPath: /csi
               name: socket-dir
--- a/examples/csi-app.yaml
+++ b/examples/csi-app.yaml
@@ -3,16 +3,6 @@ apiVersion: v1
 metadata:
   name: my-csi-app
 spec:
-  affinity:
-    podAffinity:
-      requiredDuringSchedulingIgnoredDuringExecution:
-      - labelSelector:
-          matchExpressions:
-          - key: app
-            operator: In
-            values:
-            - csi-hostpathplugin
-        topologyKey: kubernetes.io/hostname
   containers:
     - name: my-frontend
       image: busybox
--- a/pkg/hostpath/controllerserver.go
+++ b/pkg/hostpath/controllerserver.go
@@ -49,12 +49,13 @@ const (
 )
 
 type controllerServer struct {
-	caps []*csi.ControllerServiceCapability
+	caps   []*csi.ControllerServiceCapability
+	nodeID string
 }
 
-func NewControllerServer(ephemeral bool) *controllerServer {
+func NewControllerServer(ephemeral bool, nodeID string) *controllerServer {
 	if ephemeral {
-		return &controllerServer{caps: getControllerServiceCapabilities(nil)}
+		return &controllerServer{caps: getControllerServiceCapabilities(nil), nodeID: nodeID}
 	}
 	return &controllerServer{
 		caps: getControllerServiceCapabilities(
@@ -64,6 +65,7 @@ func NewControllerServer(ephemeral bool) *controllerServer {
 				csi.ControllerServiceCapability_RPC_LIST_SNAPSHOTS,
 				csi.ControllerServiceCapability_RPC_CLONE_VOLUME,
 			}),
+		nodeID: nodeID,
 	}
 }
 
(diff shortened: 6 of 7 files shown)


pkg/hostpath/identityserver.go
```diff
@@ -68,6 +68,13 @@ func (ids *identityServer) GetPluginCapabilities(ctx context.Context, req *csi.G
 					},
 				},
 			},
+			{
+				Type: &csi.PluginCapability_Service_{
+					Service: &csi.PluginCapability_Service{
+						Type: csi.PluginCapability_Service_VOLUME_ACCESSIBILITY_CONSTRAINTS,
```
msau42: `CreateVolume` also needs to be updated to return topology in the response

msau42: topology also has to be feature-gated on in the csi-provisioner spec

mucahitkurt: fixed, thanks!

msau42: sorry, I just realized that we can't update the deployment spec until we tag a new image.  So we may need to split this into 2 PRs: 

1) add topology to the driver. this may need to be manually tested with private image.  After this is merged, then we can cut a new hostpath driver image.  
2) update the deployment spec

mucahitkurt: I reverted the deployment spec changes, I'll try to test the driver on my local env.

mucahitkurt: I'm just trying to understand, why should we separate deployment spec changes and driver topology support changes?

msau42: Actually, maybe it won't cause issues.  The pull jobs should be replacing hostpath driver image with a custom one built from the PR.  But the CI jobs run with the hostpath driver image specified in the deployment spec, so they will start running with topology enabled in the provisioner, but the tagged driver doesn't have it enabled yet.  I think that may actually be fine though because the provisioner checks the feature gate + if the driver supports the capability. So actually, maybe there isn't a problem.

mucahitkurt: Then I'll add the feature gate to the provisioner?

mucahitkurt:
@msau42 I can run csi-hostpathplugin statefulset on my local one-node kubernetes 1.15 cluster without any error with the topology changes(with a local image), but when I check the `pull-kubernetes-csi-csi-driver-host-path-1-15-on-kubernetes-1-15` build logs I see hostpathplugin is not running and in a error state, there is a container error; `Container node-driver-registrar ready: false, restart count 7`, I think this the reason why tests failed. I couldn't figure out the reason, do you have any idea? Thanks!

mucahitkurt:
I'm trying to understand why these tests failed, I want to run the some of failed [test on my local kubernetes](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-testing/e2e-tests.md#local-clusters) with the changed csi hostpath driver but haven't succeeded yet.

msau42:
It looks like the tests are failing during hostpath driver bringup (node-driver-registrar is crashing?)  You can try running deploy/kubernetes-1.15/deploy-hostpath.sh on your local cluster with a private hostpath driver image

mucahitkurt:
> It looks like the tests are failing during hostpath driver bringup (node-driver-registrar is crashing?) You can try running deploy/kubernetes-1.15/deploy-hostpath.sh on your local cluster with a private hostpath driver image

For the tests `pull-kubernetes-csi-csi-driver-host-path-1-15-on-kubernetes-1-15`, I think test failed after the hostpath driver bring up because I can see hostpath driver is up before the tests are started from the [build log](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_csi-driver-host-path/88/pull-kubernetes-csi-csi-driver-host-path-1-15-on-kubernetes-1-15/1165389466169249793/build-log.txt).

But It seems that volume provisioning, attachment are successful, but mount operation failed during the tests are running, and there are events about node-driver crash.

I'm trying a similar approach on my local env., but couldn't create the same test results. I'm running csi hostpath plugin from a local image that I build, I can run [this test](https://gist.github.com/mucahitkurt/0abc0e304abc22448c77664b84d90bcc) successfully.

When I try to run a e2e test, I can get success like [this one](https://gist.github.com/mucahitkurt/0abc0e304abc22448c77664b84d90bcc).

Is there any way to get more detail(logs) about why a pod is crashing from these CSI tests?

msau42:
You can get driver logs at: https://gcsweb.k8s.io/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_csi-driver-host-path/88/pull-kubernetes-csi-csi-driver-host-path-1-15-on-kubernetes-1-15/1165389466169249793/artifacts/

msau42:
This is the error from node-driver-registrar:
```
I0904 21:35:31.590723       1 main.go:87] Received NotifyRegistrationStatus call: &RegistrationStatus{PluginRegistered:false,Error:RegisterPlugin error -- plugin registration failed with err: error updating Node object with CSI driver node info: error updating node: timed out waiting for the condition; caused by: failed to patch status "{\"metadata\":{\"annotations\":{\"csi.volume.kubernetes.io/nodeid\":\"{\\\"hostpath.csi.k8s.io\\\":\\\"csi-prow-control-plane\\\"}\"},\"labels\":{\"node.kubernetes.io/topology.hostpath.csi.k8s.io/node\":\"csi-prow-control-plane\"}}}" for node "csi-prow-control-plane": Node "csi-prow-control-plane" is invalid: metadata.labels: Invalid value: "node.kubernetes.io/topology.hostpath.csi.k8s.io/node": a qualified name must consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyName',  or 'my.name',  or '123-abc', regex used for validation is '([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]') with an optional DNS subdomain prefix and '/' (e.g. 'example.com/MyName'),}
E0904 21:35:31.591022       1 main.go:89] Registration process failed with error: RegisterPlugin error -- plugin registration failed with err: error updating Node object with CSI driver node info: error updating node: timed out waiting for the condition; caused by: failed to patch status "{\"metadata\":{\"annotations\":{\"csi.volume.kubernetes.io/nodeid\":\"{\\\"hostpath.csi.k8s.io\\\":\\\"csi-prow-control-plane\\\"}\"},\"labels\":{\"node.kubernetes.io/topology.hostpath.csi.k8s.io/node\":\"csi-prow-control-plane\"}}}" for node "csi-prow-control-plane": Node "csi-prow-control-plane" is invalid: metadata.labels: Invalid value: "node.kubernetes.io/topology.hostpath.csi.k8s.io/node": a qualified name must consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyName',  or 'my.name',  or '123-abc', regex used for validation is '([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]') with an optional DNS subdomain prefix and '/' (e.g. 'example.com/MyName'), restarting registration container.
```

pkg/hostpath/nodeserver.go
```diff
@@ -31,6 +31,8 @@ import (
 	"k8s.io/kubernetes/pkg/volume/util/volumepathhandler"
 )
 
+const TopologyKeyNode = "node.kubernetes.io/topology.hostpath.csi.k8s.io/node"
```
msau42: I would remove the "node.kubernetes.io/" prefix. I think that's what the error is saying.

mucahitkurt: It was like that but I added that prefix because my previous error message said it's forbidden to use that kind of label, so when I check the error message I found [this check](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#noderestriction), and allowed prefixes.

I think current error message is because of usage two `/` characters, so I expect that if I change the label name to something like `node.kubernetes.io/topology.hostpath.csi.k8s.io-node` it'll pass.

Do you know a way that I can over forbidden check and use node label like `topology.hostpath.csi.k8s.io/node`?

```
I0824 22:30:52.401362       1 main.go:87] Received NotifyRegistrationStatus call: &RegistrationStatus{PluginRegistered:false,Error:RegisterPlugin error -- plugin registration failed with err: error updating Node object with CSI driver node info: error updating node: timed out waiting for the condition; caused by: failed to patch status "{\"metadata\":{\"annotations\":{\"csi.volume.kubernetes.io/nodeid\":\"{\\\"hostpath.csi.k8s.io\\\":\\\"csi-prow-control-plane\\\"}\"},\"labels\":{\"topology.hostpath.csi.k8s.io/node\":\"csi-prow-control-plane\"}}}" for node "csi-prow-control-plane": nodes "csi-prow-control-plane" is forbidden: is not allowed to modify labels: topology.hostpath.csi.k8s.io/node,}
E0824 22:30:52.401510       1 main.go:89] Registration process failed with error: RegisterPlugin error -- plugin registration failed with err: error updating Node object with CSI driver node info: error updating node: timed out waiting for the condition; caused by: failed to patch status "{\"metadata\":{\"annotations\":{\"csi.volume.kubernetes.io/nodeid\":\"{\\\"hostpath.csi.k8s.io\\\":\\\"csi-prow-control-plane\\\"}\"},\"labels\":{\"topology.hostpath.csi.k8s.io/node\":\"csi-prow-control-plane\"}}}" for node "csi-prow-control-plane": nodes "csi-prow-control-plane" is forbidden: is not allowed to modify labels: topology.hostpath.csi.k8s.io/node, restarting registration container.
```

msau42: I believe the restriction is on labels with k8s.io or kubernetes.io. Try leaving that part out of the label and see if that works

mucahitkurt: I changed to `topology.hostpath.csi/node`, it works, thanks!

msau42:
Can you also update the example pods under the examples/ directory?  They currently have a nodeSelector to be on the same node where the hostpath driver is deployed, but technically with topology enabled, we don't need that anymore.

mucahitkurt:
> Can you also update the example pods under the examples/ directory? They currently have a nodeSelector to be on the same node where the hostpath driver is deployed, but technically with topology enabled, we don't need that anymore.

Only `examples/csi-app.yaml` has pod affinity settings, I deleted it.

Also for [the below comment in the issue](https://github.com/kubernetes-csi/csi-driver-host-path/issues/54#issue-449196899), should I open an issue?

```
...Then applications no longer need to be scheduled manually onto the right node and we can remove that workaround from the E2E testing in Kubernetes....
```

msau42:
/lgtm
Thanks! For the in-tree hostpath e2e changes, opening an new issue sounds good. We can update in-tree tests once we cut a new release of the driver.

PR #90, opened 2019-08-22
Update manifests for expand support

### LINKED ISSUE #86: Add support for volume resizing.
In preparation for volume resizing going beta, we should add support for it in hostpath driver and make sure the tests run with it.

bertinatto:
/kind feature

This PR updates the manifest files and deploy script in order to have volume expansion support.

This is a continuation of #87.

Fixes #86 

```release-note
Added volume expansion support.
```

### DIFF
--- a/deploy/util/deploy-hostpath.sh
+++ b/deploy/util/deploy-hostpath.sh
@@ -5,7 +5,7 @@
 # authoritative and all updates for this process should be
 # done here and referenced elsewhere.
 
-# The script assumes that kubectl is available on the OS path 
+# The script assumes that kubectl is available on the OS path
 # where it is executed.
 
 set -e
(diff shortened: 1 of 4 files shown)


gnufied:
Can we double check the script:

```
error: unable to read URL "https://raw.githubusercontent.com/kubernetes-csi/external-resizer//deploy/kubernetes/rbac.yaml", server reported 404 Not Found, status code=404
```

bertinatto:
CC @gnufied

deploy/util/deploy-hostpath.sh
```diff
@@ -147,7 +149,7 @@ done
 # Wait until all pods are running. We have to make some assumptions
 # about the deployment here, otherwise we wouldn't know what to wait
 # for: the expectation is that we run attacher, provisioner,
-# snapshotter, socat and hostpath plugin in the default namespace.
+# snapshotter, resizer, socat and hostpath plugin in the default namespace.
 cnt=0
 while [ $(kubectl get pods 2>/dev/null | grep '^csi-hostpath.* Running ' | wc -l) -lt 5 ] || ! kubectl describe volumesnapshotclasses.snapshot.storage.k8s.io 2>/dev/null >/dev/null; do
```
pohly: I think this passed testing accidentally: it adds one additional pod (external-resizer), but then doesn't bump up the number of pods that the script waits for. This works if the script sees five running pods, but if only sees four and then six, the script waits until it times out.

That would explain https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-csi-canary-on-kubernetes-1-14/1191339260892942337 which doesn't have any obvious pod failures.

PR #95, opened 2019-09-07
Update hostpath driver to latest tag in deployments

msau42:
**What type of PR is this?**
/kind feature

So that we can start testing new features added to hostpath in kubernetes-csi like topology, expansion, cloning

Fixes #

```release-note
NONE
```

### DIFF
--- a/deploy/kubernetes-1.13/hostpath/csi-hostpath-plugin.yaml
+++ b/deploy/kubernetes-1.13/hostpath/csi-hostpath-plugin.yaml
@@ -63,7 +63,7 @@ spec:
             name: csi-data-dir
 
         - name: hostpath
-          image: quay.io/k8scsi/hostpathplugin:v1.1.0
+          image: quay.io/k8scsi/hostpathplugin:v1.2.0-rc6
           args:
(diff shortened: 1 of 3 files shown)


msau42:
Opening https://github.com/kubernetes-csi/csi-driver-host-path/issues/96 for test failure.

/retest

deploy/kubernetes-1.13/hostpath/csi-hostpath-plugin.yaml
```diff
@@ -63,7 +63,7 @@ spec:
             name: csi-data-dir
 
         - name: hostpath
-          image: quay.io/k8scsi/hostpathplugin:v1.1.0
```
davidz627: do we not want to continue to "verify" `v1.1.0` compatibility with `k8s 1.13`? I guess what if a patch in `1.13` breaks compat?

msau42: At least for hostpath driver, we have not been maintaining older releases.  Patches in K8s that break CSI are tested as part of K8s in-tree test configs, which will continue pulling older versions of the hostpath driver.

PR #97, opened 2019-09-13
Kubernetes 1.16 + ephemeral inline volumes

pohly:
**What type of PR is this?**
/kind feature

For Kubernetes 1.16 a new deployment is needed with a new
csi-hostpath-driverinfo.yaml file that enables "ephemeral"
volumes. The only difference compared to the kubernetes-1.15
deployment is that file and the updated README.md.

With that deployment and Kubernetes >= 1.16, using inline volumes no
longer depends on modifying the deployment.

```release-note
ephemeral inline volumes in addition to normal volumes on Kubernetes 1.16
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -343,10 +343,10 @@ Follow the following example to create a volume from a volume snapshot:
 > ```
 
 ## Inline ephemeral support
-As of version 1.15 of Kubernetes, the CSI Hostpath driver (starting with version 1.0.1) now includes support for inline ephemeral volume.  This means that a volume can be specified directly inside a pod spec without the need to use a persistent volume object.
-Find out how to enable or create a CSI inline driver [here](https://kubernetes-csi.github.io/docs/ephemeral-local-volumes.html)
+The CSI Hostpath driver (starting with version 1.2.0) now includes support for inline ephemeral volumes. This means that a volume can be specified directly inside a pod spec without the need to use a persistent volume object.
+Find out how to enable or create a CSI driver with support for such volumes [here](https://kubernetes-csi.github.io/docs/ephemeral-local-volumes.html)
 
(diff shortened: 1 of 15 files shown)


README.md
```diff
@@ -343,10 +343,10 @@ Follow the following example to create a volume from a volume snapshot:
 > ```
 
 ## Inline ephemeral support
-As of version 1.15 of Kubernetes, the CSI Hostpath driver (starting with version 1.0.1) now includes support for inline ephemeral volume.  This means that a volume can be specified directly inside a pod spec without the need to use a persistent volume object.
-Find out how to enable or create a CSI inline driver [here](https://kubernetes-csi.github.io/docs/ephemeral-local-volumes.html)
+The CSI Hostpath driver (starting with version 1.2.0) now includes support for inline ephemeral volumes. This means that a volume can be specified directly inside a pod spec without the need to use a persistent volume object.
```
pohly: The dependency on Kubernetes is mentioned below. What matters here is the driver release which has the support. I don't think we ever released 1.0.1, did we?

1.2.0 will be the first "real" release with any kind of ephemeral support.

pohly:
The corresponding kubernetes-csi.github.io update is here: https://github.com/kubernetes-csi/docs/pull/209

deploy/kubernetes-1.16/hostpath/csi-hostpath-plugin.yaml
```diff
@@ -0,0 +1,138 @@
...
+          volumeMounts:
+          - mountPath: /csi
+            name: socket-dir
+          - mountPath: /registration
+            name: registration-dir
+          - mountPath: /csi-data-dir
+            name: csi-data-dir
+
+        - name: hostpath
+          image: quay.io/k8scsi/hostpathplugin:v1.2.0-rc6
```
msau42: Can you use rc8 which also includes the resizing stuff too?

pohly: Just here or also in the other deployments?

pohly: I bumped the version in all deployments.

PR #98, opened 2019-09-18
Add snapshot beta CRD deployment for 1.17

xing-yang:
**What type of PR is this?**
/kind feature

This PR adds snapshot beta CRD deployment for 1.17.

Fixes #

Note: This PR is not ready for review.

```release-note
Adds snapshot beta CRD deployment for 1.17
```


deploy/kubernetes-1.17/snapshotter/csi-hostpath-snapshotclass.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: snapshot.storage.k8s.io/v1alpha1
+kind: VolumeSnapshotClass
+metadata:
+  name: csi-hostpath-snapclass
+snapshotter: hostpath.csi.k8s.io
```
xing-yang: apiVersion: snapshot.storage.k8s.io/v1beta1
kind: VolumeSnapshotClass
metadata:
  name: csi-hostpath-snapclass
driver: hostpath.csi.k8s.io
deletionPolicy: Delete

xing-yang: I wonder if we still need to pre-create snapshotclass with the new apis.

xing-yang:
storageclass.yaml:

```
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: csi-hostpath-sc
provisioner: hostpath.csi.k8s.io
reclaimPolicy: Delete
volumeBindingMode: Immediate
```

snapshot.yaml
```
apiVersion: snapshot.storage.k8s.io/v1beta1
kind: VolumeSnapshot
metadata:
  name: new-snapshot-demo
spec:
  volumeSnapshotClassName: csi-hostpath-snapclass
  source:
    persistentVolumeClaimName: hpvc
```

xing-yang:
We also need to link kubernetes-latest to the 1.17 directory:
https://github.com/kubernetes-csi/csi-driver-host-path/blob/master/deploy/kubernetes-latest

deploy/kubernetes-1.17/snapshotter/csi-hostpath-snapshotclass.yaml
ggriffiths: Removed this

deploy/kubernetes-1.17/deploy-hostpath.sh
```diff
@@ -0,0 +1,172 @@
...
+: ${CSI_PROVISIONER_RBAC:=https://raw.githubusercontent.com/kubernetes-csi/external-provisioner/$(rbac_version "${BASE_DIR}/hostpath/csi-hostpath-provisioner.yaml" csi-provisioner "${UPDATE_RBAC_RULES}")/deploy/kubernetes/rbac.yaml}
+CSI_ATTACHER_RBAC_YAML="https://raw.githubusercontent.com/kubernetes-csi/external-attacher/$(rbac_version "${BASE_DIR}/hostpath/csi-hostpath-attacher.yaml" csi-attacher false)/deploy/kubernetes/rbac.yaml"
+: ${CSI_ATTACHER_RBAC:=https://raw.githubusercontent.com/kubernetes-csi/external-attacher/$(rbac_version "${BASE_DIR}/hostpath/csi-hostpath-attacher.yaml" csi-attacher "${UPDATE_RBAC_RULES}")/deploy/kubernetes/rbac.yaml}
+# TODO: Change back to dynamic path after image is released officially
+CSI_SNAPSHOTTER_RBAC_YAML="https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/master/deploy/kubernetes/rbac.yaml"
+: ${CSI_SNAPSHOTTER_RBAC:=https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/master/deploy/kubernetes/rbac.yaml}
```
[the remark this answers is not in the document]
ggriffiths: > We also need to link kubernetes-latest to the 1.17 directory:
> https://github.com/kubernetes-csi/csi-driver-host-path/blob/master/deploy/kubernetes-latest

Updated this symlink

deploy/kubernetes-1.17/deploy-hostpath.sh
```diff
@@ -0,0 +1,210 @@
...
+#: ${CSI_PROVISIONER_RBAC:=https://raw.githubusercontent.com/kubernetes-csi/external-provisioner/$(rbac_version "${BASE_DIR}/hostpath/csi-hostpath-provisioner.yaml" csi-provisioner "${UPDATE_RBAC_RULES}")/deploy/kubernetes/rbac.yaml}
+#CSI_ATTACHER_RBAC_YAML="https://raw.githubusercontent.com/kubernetes-csi/external-attacher/$(rbac_version "${BASE_DIR}/hostpath/csi-hostpath-attacher.yaml" csi-attacher false)/deploy/kubernetes/rbac.yaml"
+#: ${CSI_ATTACHER_RBAC:=https://raw.githubusercontent.com/kubernetes-csi/external-attacher/$(rbac_version "${BASE_DIR}/hostpath/csi-hostpath-attacher.yaml" csi-attacher "${UPDATE_RBAC_RULES}")/deploy/kubernetes/rbac.yaml}
+# TODO: Change back to dynamic path after image is released officially
+#CSI_SNAPSHOTTER_RBAC_YAML="https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/master/deploy/kubernetes/rbac.yaml"
+#: ${CSI_SNAPSHOTTER_RBAC:=https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/master/deploy/kubernetes/rbac.yaml}
```
xing-yang: This one should work now.  The snapshot beta API PR is merged and its rbac.yaml is updated.

deploy/kubernetes-1.17/hostpath/csi-hostpath-snapshotter.yaml
```diff
@@ -0,0 +1,57 @@
...
+              - key: app
+                operator: In
+                values:
+                - csi-hostpathplugin
+            topologyKey: kubernetes.io/hostname
+      serviceAccount: csi-snapshotter
+      containers:
+        - name: csi-snapshotter
+          # TODO: change to official image when released
+          image: quay.io/k8scsi/csi-snapshotter:testsnapshotbeta
```
xing-yang: You can use the following image which is build from the master code:
image: quay.io/k8scsi/csi-snapshotter:canary

deploy/kubernetes-1.17/hostpath/csi-hostpath-provisioner.yaml
```diff
@@ -0,0 +1,55 @@
...
+              matchExpressions:
+              - key: app
+                operator: In
+                values:
+                - csi-hostpathplugin
+            topologyKey: kubernetes.io/hostname
+      serviceAccountName: csi-provisioner
+      containers:
+        - name: csi-provisioner
+          image: quay.io/k8scsi/csi-provisioner:testsnapshotbeta #v1.4.0-rc1
```
xing-yang: Please use the following image which uses the merged snapshot beta APIs:
quay.io/k8scsi/csi-provisioner:testsplitcontroller

deploy/kubernetes-1.17/snapshotter/crd/snapshot.storage.k8s.io_volumesnapshotclasses.yaml
```diff
@@ -0,0 +1,74 @@
+---
```
xing-yang: Please update these 3 CRDs using the lates ones here. They are merged into master branch of external-snapshotter now:
https://github.com/kubernetes-csi/external-snapshotter/tree/master/config/crd

deploy/kubernetes-1.17/README.md
```diff
@@ -0,0 +1,7 @@
+The deployment for Kubernetes 1.15 uses CSI 1.0 and thus is
+incompatible with Kubernetes < 1.13.
+
+The sidecars depend on 1.15 API changes for migration and resizing,
+and 1.14 API changes for CSIDriver and CSINode.
+However the hostpath driver doesn't use those features, so this
+deployment can work on older Kubernetes versions.
```
xing-yang: This Readme does not apply any more.  Replace this readme with following:

The deployment for Kubernetes 1.17 installs VolumeSnapshot Beta CRDs and thus is imcompatible
with Kubernetes < 1.17 when the VolumeSnapshot CRDs were Alpha.

deploy/kubernetes-1.17/deploy-hostpath.sh
ggriffiths: Updated this

deploy/kubernetes-1.17/snapshotter/crd/snapshot.storage.k8s.io_volumesnapshotclasses.yaml
ggriffiths: Updated with the lastest CRDs

deploy/kubernetes-1.17/deploy-hostpath.sh
```diff
@@ -0,0 +1,203 @@
...
+
+    echo "$version"
+}
+
+# In addition, the RBAC rules can be overridden separately.
+#
+#CSI_PROVISIONER_RBAC_YAML="https://raw.githubusercontent.com/kubernetes-csi/external-provisioner/$(rbac_version "${BASE_DIR}/hostpath/csi-hostpath-provisioner.yaml" csi-provisioner false)/deploy/kubernetes/rbac.yaml"
+#: ${CSI_PROVISIONER_RBAC:=https://raw.githubusercontent.com/kubernetes-csi/external-provisioner/$(rbac_version "${BASE_DIR}/hostpath/csi-hostpath-provisioner.yaml" csi-provisioner "${UPDATE_RBAC_RULES}")/deploy/kubernetes/rbac.yaml}
+#CSI_ATTACHER_RBAC_YAML="https://raw.githubusercontent.com/kubernetes-csi/external-attacher/$(rbac_version "${BASE_DIR}/hostpath/csi-hostpath-attacher.yaml" csi-attacher false)/deploy/kubernetes/rbac.yaml"
+#: ${CSI_ATTACHER_RBAC:=https://raw.githubusercontent.com/kubernetes-csi/external-attacher/$(rbac_version "${BASE_DIR}/hostpath/csi-hostpath-attacher.yaml" csi-attacher "${UPDATE_RBAC_RULES}")/deploy/kubernetes/rbac.yaml}
```
xing-yang: We didn't change these RBAC's.  These should just work?

deploy/kubernetes-1.17/deploy-hostpath.sh
```diff
@@ -0,0 +1,203 @@
...
+# TODO: Change back to dynamic path after image is released officially
+CSI_SNAPSHOTTER_RBAC_YAML="https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/master/deploy/kubernetes/rbac.yaml"
+: ${CSI_SNAPSHOTTER_RBAC:=https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/master/deploy/kubernetes/rbac.yaml}
+
+# Using temporary rbac yaml files
+CSI_PROVISIONER_RBAC_YAML="https://raw.githubusercontent.com/kubernetes-csi/external-provisioner/master/deploy/kubernetes/rbac.yaml"
+: ${CSI_PROVISIONER_RBAC:=https://raw.githubusercontent.com/kubernetes-csi/external-provisioner/master/deploy/kubernetes/rbac.yaml}
+CSI_ATTACHER_RBAC_YAML="https://raw.githubusercontent.com/kubernetes-csi/external-attacher/master/deploy/kubernetes/rbac.yaml"
+: ${CSI_ATTACHER_RBAC:=https://raw.githubusercontent.com/kubernetes-csi/external-attacher/master/deploy/kubernetes/rbac.yaml}
```
xing-yang: I don't think we need these temporary rbac files any more.

deploy/kubernetes-1.17/hostpath/csi-hostpath-snapshotter.yaml
```diff
@@ -0,0 +1,57 @@
...
+      serviceAccount: csi-snapshotter
+      containers:
+        - name: csi-snapshotter
+          # TODO: change to official image when released
+          image: quay.io/k8scsi/csi-snapshotter:canary
+          imagePullPolicy: IfNotPresent
+          args:
+            - -v=5
+            - --csi-address=/csi/csi.sock
+            - --connection-timeout=15s
```
xing-yang: This flag is deprecated.

deploy/kubernetes-1.17/deploy-hostpath.sh
ggriffiths: Remove these

deploy/kubernetes-1.17/hostpath/csi-hostpath-snapshotter.yaml
ggriffiths: Ok, removed

deploy/kubernetes-1.17/hostpath/csi-hostpath-attacher.yaml
```diff
@@ -0,0 +1,55 @@
...
+              matchExpressions:
+              - key: app
+                operator: In
+                values:
+                - csi-hostpathplugin
+            topologyKey: kubernetes.io/hostname
+      serviceAccountName: csi-attacher
+      containers:
+        - name: csi-attacher
+          image: quay.io/k8scsi/csi-attacher:v1.2.0
```
xing-yang: Most of the images have already been updated to a higher version in deploy/kubernetes-1.16.  For example: 
https://github.com/kubernetes-csi/csi-driver-host-path/blob/master/deploy/kubernetes-1.16/hostpath/csi-hostpath-attacher.yaml#L43

Can you check all of them and update accordingly?

deploy/kubernetes-1.17/deploy-hostpath.sh
```diff
@@ -0,0 +1,196 @@
...
+    sleep 10
+  done
+}
+
+run () {
+    echo "$@" >&2
+    "$@"
+}
+
+# deploy volume snapshot CRDs
```
msau42: we are deploying the crds twice?

Also it would be good install the CRDs directly from the snapshotter repo instead of copying them here.

ggriffiths: Installing from the master snapshotter repo now. Also removed the extra CRD deploy.

deploy/kubernetes-1.17/deploy-hostpath.sh
```diff
@@ -117,10 +118,6 @@ run () {
     "$@"
 }
 
-# deploy volume snapshot CRDs
-echo "deploying volume snapshot CRDs"
-kubectl apply -f ${BASE_DIR}/snapshotter/crd
-
 # wait until CRDs are ready
 echo "waiting for CRDs to be ready"
 create-crds-and-wait-for-apiserver
```
msau42: Should we gate this on the INSTALL_CRDS env since it may already be installed in some environments?

deploy/kubernetes-1.17/deploy-hostpath.sh
```diff
@@ -93,19 +93,20 @@ update_image () {
 # not be available at this point. Run this as a background process.
 function create-crds-and-wait-for-apiserver {
   # Wait until volumesnapshot CRDs are in place.
-  kubectl apply -f ${BASE_DIR}/snapshotter/crd/snapshot.storage.k8s.io_volumesnapshotclasses.yaml
+  CRD_BASE_DIR=https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/master/config/crd
```
msau42: We may want to make this variable configurable so that that we can have CI jobs use fixed versions of the rbacs from specific releases.

ggriffiths: I just made this base dir depend on the snapshotter version provided in `hostpath/csi-hostpath-snapshotter.yaml`. This is also how we get the correct RBAC yamls to apply. 

Also used the INSTALL_CRD flag. It defaults to true, unless user input is provided to that flag.

deploy/kubernetes-1.17/deploy-hostpath.sh
```diff
@@ -68,18 +68,19 @@ function sidecar_version () {
     # canary images were built from the corresponding branch.
     case "$version" in canary) version=master;;
                         *-canary) version="$(echo "$version" | sed -e 's/\(.*\)-canary/release-\1/')";;
+						*) version=master;;
```
ggriffiths: This is to handle the custom sidecar images we're temporarily referring to.

xing-yang:
We should probably focus on getting the split controller PR merged first then work on this one.  I just realized that this PR is using the snapshotter canary image so it does not handle the installation of the common snapshot controller yet.

deploy/kubernetes-1.17/deploy-hostpath.sh
```diff
@@ -0,0 +1,200 @@
...
+: ${CSI_PROVISIONER_RBAC:=https://raw.githubusercontent.com/kubernetes-csi/external-provisioner/$(sidecar_version "${BASE_DIR}/hostpath/csi-hostpath-provisioner.yaml" csi-provisioner "${UPDATE_RBAC_RULES}")/deploy/kubernetes/rbac.yaml}
+CSI_ATTACHER_RBAC_YAML="https://raw.githubusercontent.com/kubernetes-csi/external-attacher/$(sidecar_version "${BASE_DIR}/hostpath/csi-hostpath-attacher.yaml" csi-attacher false)/deploy/kubernetes/rbac.yaml"
+: ${CSI_ATTACHER_RBAC:=https://raw.githubusercontent.com/kubernetes-csi/external-attacher/$(sidecar_version "${BASE_DIR}/hostpath/csi-hostpath-attacher.yaml" csi-attacher "${UPDATE_RBAC_RULES}")/deploy/kubernetes/rbac.yaml}
+CSI_SNAPSHOTTER_RBAC_YAML="https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/$(sidecar_version "${BASE_DIR}/hostpath/csi-hostpath-snapshotter.yaml" csi-snapshotter "${UPDATE_RBAC_RULES}")/deploy/kubernetes/rbac.yaml"
+: ${CSI_SNAPSHOTTER_RBAC:=https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/$(sidecar_version "${BASE_DIR}/hostpath/csi-hostpath-snapshotter.yaml" csi-snapshotter "${UPDATE_RBAC_RULES}")/deploy/kubernetes/rbac.yaml}
+
+INSTALL_CRD=${INSTALL_CRD:-"true"}
```
xing-yang: @ggriffiths Can you submit a PR to csi-release-tools and move this INSTALL_CRD step to prow.sh after the cluster is installed?

https://github.com/kubernetes-csi/csi-release-tools/blob/master/prow.sh#L989

In additional to install CRDs, please also add the "intalling snapshot_controller" step in prow.sh, after installing the CRDs.

yaml file for installing snapshot controller (the common controller) is here:
 https://github.com/kubernetes-csi/external-snapshotter/deploy/kubernetes/snapshot-controller

You'll need to wait for CRDs and controller to be installed.  See my PR here for example: 
https://github.com/kubernetes/kubernetes/pull/80058/files#diff-6e3b476b9225d1213dc6ad13e453fc16R2081

The sidecar controller installation should still be in csi-driver-host-path deploy-hostpath.sh.
https://github.com/kubernetes-csi/external-snapshotter/deploy/kubernetes/csi-snapshotter

xing-yang: by the way, the Dockerfile's are under external-snapshotter/cmd if you need to build images yourself for some reason.

pohly:
> @ggriffiths Can you submit a PR to csi-release-tools and move this INSTALL_CRD step to prow.sh after the cluster is installed?

The usual approach for this is:
- modify `release-tools/prow.sh` in this PR with a separate commit
- update your branch
- check that it works in Prow while ignoring the unit test failure about locally modified `release-tools`
- submit to `csi-release-tools`
- delete the separate commit and update `release-tools` via `git submodule`

`release-tools/README.md` shows how to work with `git submodule`. There's a command for submitting the local commit upstream.

> yaml file for installing snapshot controller (the common controller) is here

This leads to an "interesting" challenge: which version should be installed? In most jobs it has to be a fixed, deterministic version, probably the version intended for a certain Kubernetes release. But what about testing changes to that code? Where will those changes be tested? We'll need jobs that cover the various version skew scenarios... can be dealt with later, of course.

xing-yang:
@msau42 See comments from @pohly.  I wonder if we should just install snapshot CRDs and snapshot controller using hostpath driver's deployment's script?  This way it is easy to determine what version to install because each folder, i.e., 1.17, will be installing a specific version.  Thoughts?

pohly:
Xing Yang <notifications@github.com> writes:
> @msau42 See comments from @pohly.  I wonder if we should just install
> snapshot CRDs and snapshot controller using hostpath driver's
> deployment's script?

No, I agree with @msau42 that this CRD is something that belongs to the
cluster and not one particular CSI driver. Otherwise we have two
different CSI drivers fighting over who gets to install "his" version of
the CRD.

> This way it is easy to determine what version to install because each
> folder, i.e., 1.17, will be installing a specific version.  Thoughts?

That doesn't address the underlying version skew issue. In practice it
can still occur once a cluster has more than one CSI driver.

deploy/kubernetes-1.17/hostpath/csi-hostpath-provisioner.yaml
```diff
@@ -0,0 +1,54 @@
...
+              matchExpressions:
+              - key: app
+                operator: In
+                values:
+                - csi-hostpathplugin
+            topologyKey: kubernetes.io/hostname
+      serviceAccountName: csi-provisioner
+      containers:
+        - name: csi-provisioner
+          image: quay.io/k8scsi/csi-provisioner:testsplitcontroller
```
xing-yang: This should use quay.io/k8scsi/csi-provisioner:v1.5.0-rc1

deploy/kubernetes-1.17/hostpath/csi-hostpath-snapshotter.yaml
```diff
@@ -0,0 +1,56 @@
...
+              - key: app
+                operator: In
+                values:
+                - csi-hostpathplugin
+            topologyKey: kubernetes.io/hostname
+      serviceAccount: csi-snapshotter
+      containers:
+        - name: csi-snapshotter
+          # TODO: change to official image when released
+          image: quay.io/k8scsi/csi-snapshotter:canary
```
xing-yang: Use quay.io/k8scsi/csi-snapshotter:v2.0.0-rc2

xing-yang:
Are you installing snapshot-controller any where?  I could't find it.

quay.io/k8scsi/snapshot-controller:v2.0.0-rc2

ggriffiths:
> Are you installing snapshot-controller any where? I could't find it.
> 
> quay.io/k8scsi/snapshot-controller:v2.0.0-rc2

Added this installation (along with the RBAC yaml for the snapshot controller)

ggriffiths:
Added snapshot-controller and updated release-tools, but now some tests are failing. Will fix this next.

ggriffiths:
@xing-yang This is ready for review again. I've squashed all intermediate commits into one temporary prow.sh commit and another commit with my other changes.

The prow.sh changes are also in this release-tool PR:
https://github.com/kubernetes-csi/csi-release-tools/pull/45

release-tools/prow.sh
```diff
@@ -657,6 +657,29 @@ install_hostpath () {
     fi
 }
 
+# Installs all nessesary snapshotter CRDs  
+install_snapshot_crds() {
+  # Wait until volumesnapshot CRDs are in place.
+  CRD_BASE_DIR=https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/master/config/crd
```
pohly: One key aspect of a Prow test job is that its input is fixed and deterministic. Using `master` here makes it non-deterministic which CRD gets installed (will change over time).

Can you replace with a tagged version or a commit hash?

release-tools/prow.sh
```diff
@@ -657,6 +657,29 @@ install_hostpath () {
     fi
 }
 
+# Installs all nessesary snapshotter CRDs  
+install_snapshot_crds() {
+  # Wait until volumesnapshot CRDs are in place.
+  CRD_BASE_DIR=https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/master/config/crd 
+  kubectl apply -f "${CRD_BASE_DIR}/snapshot.storage.k8s.io_volumesnapshotclasses.yaml"
+  until kubectl get volumesnapshotclasses.snapshot.storage.k8s.io
```
pohly: Perhaps move all the wait loops to the end, i.e. first create CRDs, let them be processed in parallel, then wait?

release-tools/prow.sh
```diff
@@ -987,6 +1010,9 @@ main () {
         if tests_need_non_alpha_cluster; then
             start_cluster || die "starting the non-alpha cluster failed"
 
+            # Install nessesary CRDs
```
pohly: Typo: necessary

release-tools/prow.sh
xing-yang: @ggriffiths you can use this tag: https://github.com/kubernetes-csi/external-snapshotter/tree/v2.0.0-rc2/config/crd

release-tools/prow.sh
```diff
@@ -987,6 +1010,9 @@ main () {
...
+            install_snapshot_crds
+
```
xing-yang: deploy/kubernetes/snapshot-controller/rbac-snapshot-controller.yaml and deploy/kubernetes/snapshot-controller/setup-snapshot-controller.yaml also should be installed in prow.sh, after the CRD install

deploy/kubernetes-1.17/deploy-hostpath.sh
```diff
@@ -0,0 +1,169 @@
...
+CSI_SNAPSHOTTER_RBAC_YAML="https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/$(sidecar_version "${BASE_DIR}/hostpath/csi-hostpath-snapshotter.yaml" csi-snapshotter "${UPDATE_RBAC_RULES}")/deploy/kubernetes/csi-snapshotter/rbac-csi-snapshotter.yaml"
+: ${CSI_SNAPSHOTTER_RBAC:=https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/$(sidecar_version "${BASE_DIR}/hostpath/csi-hostpath-snapshotter.yaml" csi-snapshotter "${UPDATE_RBAC_RULES}")/deploy/kubernetes/csi-snapshotter/rbac-csi-snapshotter.yaml}
+CSI_SNAPSHOT_CONTROLLER_RBAC_YAML="https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/$(sidecar_version "${BASE_DIR}/hostpath/csi-hostpath-snapshotter.yaml" csi-snapshotter "${UPDATE_RBAC_RULES}")/deploy/kubernetes/snapshot-controller/rbac-snapshot-controller.yaml"
+: ${CSI_SNAPSHOT_CONTROLLER_RBAC:=https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/$(sidecar_version "${BASE_DIR}/hostpath/csi-hostpath-snapshotter.yaml" csi-snapshotter "${UPDATE_RBAC_RULES}")/deploy/kubernetes/snapshot-controller/rbac-snapshot-controller.yaml}
```
xing-yang: snapshot-controller yaml should be moved to prow.sh.

deploy/kubernetes-1.17/hostpath/csi-hostpath-snapshot-controller.yaml
```diff
@@ -0,0 +1,27 @@
+# This YAML file shows how to deploy the snapshot controller
```
xing-yang: This should be installed in prow.sh.

release-tools/prow.sh
ggriffiths: I was actually seeing issues when the CRDs were applied all at once. Sequentially applying them seemed to fix an issue where SnapshotContent status was not showing up.

However, I'll try applying them in parallel again and see if that is still happening.

deploy/kubernetes-1.17/hostpath/csi-hostpath-snapshot-controller.yaml
ggriffiths: Moved this.

deploy/kubernetes-1.17/deploy-hostpath.sh
ggriffiths: Moved this.

release-tools/prow.sh
ggriffiths: Moved there.

release-tools/prow.sh
```diff
@@ -657,6 +660,48 @@ install_hostpath () {
...
+# Install snapshot controller and associated RBAC, retrying until the pod is running.
+install_snapshot_controller() {
+  kubectl apply -f "https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/${CSI_SNAPSHOTTER_VERSION}/deploy/kubernetes/snapshot-controller/rbac-snapshot-controller.yaml"
+  
+  kubectl apply -f "https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/${CSI_SNAPSHOTTER_VERSION}/deploy/kubernetes/snapshot-controller/setup-snapshot-controller.yaml"
+  cnt=0
+  while [ "$(kubectl get pods 2>/dev/null | grep -c '^snapshot-controller.* Running ')" -lt 1 ]; do
+    if [ $cnt -gt 30 ]; then
+        echo "Running pods:"
+        kubectl describe pods
```
xing-yang: Is this good enough to find the snapshot-controller pod?  Is it possible that there are other pods?  Can you try something like the following to be more specific?

https://github.com/kubernetes/kubernetes/blob/master/cluster/gce/gci/configure-helper.sh#L2106

release-tools/prow.sh
```diff
@@ -657,6 +660,48 @@ install_hostpath () {
...
+  until kubectl get volumesnapshotcontents.snapshot.storage.k8s.io
+  do
+    sleep 2
+  done
+}
+
+# Install snapshot controller and associated RBAC, retrying until the pod is running.
+install_snapshot_controller() {
+  kubectl apply -f "https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/${CSI_SNAPSHOTTER_VERSION}/deploy/kubernetes/snapshot-controller/rbac-snapshot-controller.yaml"
+
```
xing-yang: https://github.com/kubernetes/kubernetes/blob/master/cluster/gce/gci/configure-helper.sh#L2100

ggriffiths: Added.

release-tools/prow.sh
ggriffiths: Change to work like that check.

release-tools/prow.sh
pohly: That sounds like a bug in Kubernetes. If you can still reproduce it, then let's report the issue upstream and keep the current code with a comment that includes a link to that upstream bug.

deploy/util/deploy-hostpath.sh
```diff
@@ -184,6 +184,3 @@ while [ $(kubectl get pods 2>/dev/null | grep '^csi-hostpath.* Running ' | wc -l
     sleep 10
 done
 
-# deploy snapshotclass
-echo "deploying snapshotclass"
-kubectl apply -f ${BASE_DIR}/snapshotter/csi-hostpath-snapshotclass.yaml
```
pohly: Isn't this part still needed for Kubernetes <= 1.16?

ggriffiths: This was causing a failure in one of the prow jobs, so I removed it. I remember @xing-yang said we could remove the snapshotclass deploy, but that may have only been for 1.17.

@xing-yang is this needed for k8s <= 1.16?

xing-yang: This should still be needed for earlier versions that don't have the new CRD's installed.  I think it failed because the yaml file is no longer accurate.  The VolumeSnapshotClass spec is changed in the beta version.

ggriffiths: Ok, I'll re-add and update the yaml file.

xing-yang: I don't think you should update the yaml file.  Since this yaml file will only be used for earlier version of k8s, its spec should stay unchanged.

deploy/kubernetes-1.17/deploy-hostpath.sh
```diff
@@ -0,0 +1,169 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
ggriffiths: Ah yes, I copied it to test my changes for 1.17. Forgot to integrate the changes there.

I've moved the changes to the util deploy-hostpath. Testing them now, will need to add additional k8s version checks next.

deploy/kubernetes-1.16/snapshotter/csi-hostpath-snapshotclass.yaml
```diff
@@ -1,5 +1,6 @@
-apiVersion: snapshot.storage.k8s.io/v1alpha1
+apiVersion: snapshot.storage.k8s.io/v1beta1
```
ggriffiths: This needed to change, because we're temporarily testing with k8s 1.16 until 1.17 is GA. 1.16 prow will fail unless we update this.

deploy/util/deploy-hostpath.sh
ggriffiths: I needed to update this file for 1.16, because we're temporarily applying the CRDs for k8s >= 1.16.

deploy/kubernetes-1.17/deploy-hostpath.sh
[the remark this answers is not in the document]
ggriffiths: Moved all changes to util/deploy-hostpath.sh.

release-tools/prow.sh
ggriffiths: I actually tried applying the CRDs all at once again, and did not find that issue I originally ran into. Must have been a different problem when I saw the missing SnapshotContent status.

I've changed `install_snapshot_crd` to apply the CRDs all at once and wait at the end. I've also added a max retries, similar to install_snapshot_controller.

ggriffiths:
Addressed most recent round of feedback @xing-yang @pohly and updated the corresponding csi-release-tool PR with these changes:
https://github.com/kubernetes-csi/csi-release-tools/pull/45

Also did a round of manual testing and the 1.17 hostpath deploy is working with `install_snapshot_crds` and `install_snapshot_controller`:
```
[grant@PDC4-SM26-N8 external-snapshotter]$ kubectl describe volumesnapshot
Name:         new-snapshot-demo
Namespace:    default
Labels:       <none>
Annotations:  kubectl.kubernetes.io/last-applied-configuration:
                {"apiVersion":"snapshot.storage.k8s.io/v1beta1","kind":"VolumeSnapshot","metadata":{"annotations":{},"name":"new-snapshot-demo","namespace...
API Version:  snapshot.storage.k8s.io/v1beta1
Kind:         VolumeSnapshot
Metadata:
  Creation Timestamp:  2019-11-14T02:54:17Z
  Finalizers:
    snapshot.storage.kubernetes.io/volumesnapshot-as-source-protection
    snapshot.storage.kubernetes.io/volumesnapshot-bound-protection
  Generation:        1
  Resource Version:  1368
  Self Link:         /apis/snapshot.storage.k8s.io/v1beta1/namespaces/default/volumesnapshots/new-snapshot-demo
  UID:               12cc8c58-4330-4c7b-b976-a2504d39e705
Spec:
  Source:
    Persistent Volume Claim Name:  hpvc
  Volume Snapshot Class Name:      csi-hostpath-snapclass
Status:
  Bound Volume Snapshot Content Name:  snapcontent-12cc8c58-4330-4c7b-b976-a2504d39e705
  Creation Time:                       2019-11-14T02:54:17Z
  Ready To Use:                        true
  Restore Size:                        1Gi
Events:                                <none>
```

deploy/kubernetes-1.16/snapshotter/csi-hostpath-snapshotclass.yaml
xing-yang: Sounds good.

pohly: Can you document this with a comment in the code and file an issue about reverting it?

pohly: If this is the only remaining change, then don't bother and just file an issue as reminder.

ggriffiths:
> /test pull-kubernetes-csi-csi-driver-host-path-alpha-1-16-on-kubernetes-1-16

@pohly how does this test differ from the non-alpha 1-16 job? Want to know so that I can fix the failure.

It seems the alpha-1-16 job failed here:
```
deploying snapshotclass
error: unable to recognize "/home/prow/go/src/github.com/kubernetes-csi/csi-driver-host-path/deploy/kubernetes-1.16/snapshotter/csi-hostpath-snapshotclass.yaml": no matches for kind "VolumeSnapshotClass" in version "snapshot.storage.k8s.io/v1beta1"
```

deploy/kubernetes-1.16/snapshotter/csi-hostpath-snapshotclass.yaml
ggriffiths: I'm not sure if this is the only remaining change, will still need to vendor in the release-tools once that PR is merged for the prow.sh change.

I'll file the issue and add a comment referencing it.

pohly:
> @pohly how does this test differ from the non-alpha 1-16 job? Want to know so that I can fix the failure (if required for merging)

Alpha jobs are not normally run for pull requests. They run the tests marked with `Feature`, which may or may not work because those tests are for features that aren't beta yet: https://github.com/kubernetes/test-infra/blob/b0e9f8ba4522c3ce4bb6b102e427a9b1cfafb8ed/config/jobs/kubernetes-csi/csi-driver-host-path/csi-driver-host-path-config.yaml#L278

I was just curious whether that jobs triggers additional tests. I'm not sure whether it currently does. Was the failure you quoted from a test?

ggriffiths:
Ah I see, makes sense on the Alpha jobs.

> Was the failure you quoted from a test?

Yes, that failure was from the original `pull-kubernetes-csi-csi-driver-host-path-alpha-1-16-on-kubernetes-1-16` that you ran.

deploy/kubernetes-1.16/snapshotter/csi-hostpath-snapshotclass.yaml
pohly: I'm probably missing something, so just keep doing it, but let me ask anyway: why are you "temporarily testing with k8s 1.16 until 1.17 is GA"?

Wouldn't it be simpler to add the new jobs for the 1.17 deployment in test-infra without enabling them and then manually trigger them here?

pohly: You wouldn't be able to test "1.17 on Kubernetes 1.17" because the latter isn't ready, but "1.17 on Kubernetes master" should be close enough.

ggriffiths: That sounds like it could works too - I'm unfamiliar with how test-infra invokes tests/how to add them for 1.17.

Happy to add them in test-infra though if we want to!

msau42: You can add optional 1.17-on-X jobs here: https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-csi/gen-jobs.sh#L268

msau42:
I don't think the alpha test should fail:
/test pull-kubernetes-csi-csi-driver-host-path-alpha-1-16-on-kubernetes-1-16

ggriffiths:
The alpha test is failing because we're only applying the snapshot CRDs in the prow.sh non-alpha code path. This causes the snapshotclass apply to fail, because it's currently referencing the `v1beta1` snapshotclass API version.

I'll do the following next:
1. Add an optional 1.17-on-X job in test-infra
2. Change prow.sh snapshot CRD installation to happen for k8s 1.17+ (both alpha and non-alpha)
3. Revert changes to test 1.17 CRDs on 1.16.

This will allow 1.16 to go back to its previous behavior, and actually test the 1.17 Beta API on master/1.17.

ggriffiths:
Added the 1.17 on k8s master tests, and it seems that all tests are passing now (except for the unit test that checks for changes to prow.sh).

I opened a csi-release-tools PR with these changes here:
https://github.com/kubernetes-csi/csi-release-tools/pull/45

@pohly @xing-yang

deploy/kubernetes-1.17/README.md
```diff
@@ -0,0 +1,2 @@
+The deployment for Kubernetes 1.17 installs VolumeSnapshot Beta CRDs and thus is imcompatible
+with Kubernetes < 1.17 when the VolumeSnapshot CRDs were Alpha.
```
pohly: Should it say "uses VolumeSnapshot Beta CRDs" instead of "installs"?

I looked at the diff between the 1.16 and 1.17 deployment and one change looks odd:
```
*** deploy/kubernetes-1.16/hostpath/csi-hostpath-plugin.yaml	2019-10-28 19:58:36.708860196 +0100
--- deploy/kubernetes-1.17/hostpath/csi-hostpath-plugin.yaml	2019-11-19 08:59:19.110648152 +0100
***************
*** 102,109 ****
                name: plugins-dir
              - mountPath: /csi-data-dir
                name: csi-data-dir
!             - mountPath: /dev
!               name: dev-dir
          - name: liveness-probe
            volumeMounts:
            - mountPath: /csi
--- 102,108 ----
                name: plugins-dir
              - mountPath: /csi-data-dir
                name: csi-data-dir
! 
          - name: liveness-probe
            volumeMounts:
            - mountPath: /csi
***************
*** 137,143 ****
              path: /var/lib/csi-hostpath-data/
              type: DirectoryOrCreate
            name: csi-data-dir
-         - hostPath:
-             path: /dev
-             type: Directory
-           name: dev-dir
```

Why does the hostpath driver no longer need access to `/dev`?

xing-yang: Should still need it.  Probably just missed it when copying stuff over.

ggriffiths: Re-added this and updated the README to say "uses" instead.

ggriffiths:
Squashed and addressed recent feedback.
/test pull-kubernetes-csi-csi-driver-host-path-alpha-1-16-on-kubernetes-1-16
/test pull-kubernetes-csi-csi-driver-host-path-1-17-on-kubernetes-master

ggriffiths:
Squashed and addressed recent feedback.
/test pull-kubernetes-csi-csi-driver-host-path-alpha-1-16-on-kubernetes-1-16
/test pull-kubernetes-csi-csi-driver-host-path-1-17-on-kubernetes-master

ggriffiths:
Update prow.sh with latest changes.
/test pull-kubernetes-csi-csi-driver-host-path-1-17-on-kubernetes-master

release-tools/prow.sh
```diff
@@ -987,6 +1060,16 @@ main () {
         if tests_need_non_alpha_cluster; then
             start_cluster || die "starting the non-alpha cluster failed"
 
+            # Install necessary snapshot CRDs and snapshot controller
+            # For Kubernetes 1.17+, we will install the CRDs and snapshot controller.
+            if version_gt "${CSI_PROW_KUBERNETES_VERSION}" "1.16.255" || "${CSI_PROW_KUBERNETES_VERSION}" == "latest"; then
```
msau42: Does this work when CSI_PROW_KUBERNETES_VERSION is something like "release-1.16"? For example: https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-csi/csi-driver-host-path/csi-driver-host-path-config.yaml#L432

xing-yang: No, it won't work.  I thought CSI_PROW_KUBERNETES_VERSION will always be like "1.16.x".

xing-yang: What are all the possible formats for CSI_PROW_KUBERNETES_VERSION?  "release-1.16", "1.16.x", ... Anything else?  @msau42 @pohly

deploy/util/deploy-hostpath.sh
xing-yang: It's on line 201

release-tools/prow.sh
msau42: and "latest". See
https://github.com/kubernetes-csi/csi-driver-host-path/blob/f10a4ad59bbbbc622445e75a7c7451f8938d9856/release-tools/prow.sh#L125

deploy/util/deploy-hostpath.sh
msau42: It only deploys on deployment < 1.17 since the csi-hostpath-snapshotclass.yaml file is not part of the deployment. My question is why do we not need it for 1.17? My understanding is that we still need to define a snapshotclass for beta.

xing-yang: @ggriffiths can you check if the new snapshot class yaml file (for v1beta1) is installed any where?  I couldn't find it.  Without that, the snapshot tests won't pass though.

release-tools/prow.sh
```diff
@@ -657,6 +673,59 @@ install_hostpath () {
...
+    fi
+    echo "$(date +%H:%M:%S)" "waiting for snapshot RBAC setup complete, attempt #$cnt"
+	cnt=$((cnt + 1))
+    sleep 10   
+  done
+
+
+  kubectl apply -f "https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/${CSI_SNAPSHOTTER_VERSION}/deploy/kubernetes/snapshot-controller/setup-snapshot-controller.yaml"
+  cnt=0
+  until kubectl get statefulset snapshot-controller | grep snapshot-controller | grep "1/1"; do
```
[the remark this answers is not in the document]
ggriffiths: ok, I can update this in the release-tools repo. Would we want to get that change into this PR? Or can it wait until we update release-tools again?

deploy/util/deploy-hostpath.sh
```diff
@@ -82,13 +82,23 @@ function rbac_version () {
     echo "$version"
 }
 
+function version_gt() { 
+    test "$(printf '%s\n' "$@" | sort -V | head -n 1)" != "$1"; 
+}
+
 # In addition, the RBAC rules can be overridden separately.
+SNAPSHOTTER_RBAC_RELATIVE_PATH="rbac.yaml"
+if version_gt $(rbac_version "${BASE_DIR}/hostpath/csi-hostpath-snapshotter.yaml" csi-snapshotter "${UPDATE_RBAC_RULES}") "v1.2.2"; then
```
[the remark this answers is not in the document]
ggriffiths: Ah, I was mistaken on the snapshotter versioning. Seems like only 2.x+ will have the split controller.

release-tools/prow.sh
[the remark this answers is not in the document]
ggriffiths: Nevermind, seems like another change is needed in prow.sh anyway. Will add these changes here and open a release-tools PR afterwards.

ggriffiths:
Removed the release-tools subtree update to make updating this PR easier for now.

Also updated prow.sh with recent and feedback, and added k8s 1.17 snapshotclass apply back in.

/test pull-kubernetes-csi-csi-driver-host-path-1-17-on-kubernetes-master

release-tools/prow.sh
ggriffiths: Updated this to handle the following formats:
- release-x.y.z
- x.y.z
- vx.y.z

I just strip `release-` or `v` to normalize the comparison.

deploy/util/deploy-hostpath.sh
```diff
@@ -82,13 +82,45 @@ function rbac_version () {
...
+#   where X,Y, and Z are any number.
+#
+# Partial versions (1.2, release-1.2) work as well.
+# The follow substrings are stripped before version comparison:
+#   - "v"
+#   - "release-"
+#
+# Usage:
+# version_gt release-1.3 v1.2.0  (returns true)
+# version_gt v1.1.1 v1.2.0  (returns true)
```
msau42: this should be false?

deploy/util/deploy-hostpath.sh
```diff
@@ -82,13 +82,45 @@ function rbac_version () {
...
+# version_gt 1.1.1 v1.2.0  (returns false)
+# version_gt 1.1.1 release-1.2.0  (returns false)
+# version_gt 1.1.1 release-1.2.0  (returns false)
```
msau42: what's the difference between this and the previous line?

deploy/util/deploy-hostpath.sh
```diff
@@ -185,5 +217,10 @@ while [ $(kubectl get pods 2>/dev/null | grep '^csi-hostpath.* Running ' | wc -l
...
 
 # deploy snapshotclass
-echo "deploying snapshotclass"
-kubectl apply -f ${BASE_DIR}/snapshotter/csi-hostpath-snapshotclass.yaml
+SNAPSHOTCLASS_PATH="${BASE_DIR}/snapshotter/csi-hostpath-snapshotclass.yaml"
+if [ -f "$SNAPSHOTCLASS_PATH" ]; then
+    echo "deploying snapshotclass"
+    kubectl apply -f $SNAPSHOTCLASS_PATH
+else
+    echo "skipping snapshotclass"
```
msau42: is there a case where it won't exist?

deploy/util/deploy-hostpath.sh
```diff
@@ -82,13 +82,45 @@ function rbac_version () {
...
+#
+# Partial versions (1.2, release-1.2) work as well.
+# The follow substrings are stripped before version comparison:
+#   - "v"
+#   - "release-"
+#
+# Usage:
+# version_gt release-1.3 v1.2.0  (returns true)
+# version_gt v1.1.1 v1.2.0  (returns true)
+# version_gt 1.1.1 v1.2.0  (returns false)
```
msau42: what about something like 1.2 and 1.2.2?

release-tools/prow.sh
```diff
@@ -657,6 +660,59 @@ install_hostpath () {
...
+    sleep 10   
+  done
+
+
+  kubectl apply -f "https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/${CSI_SNAPSHOTTER_VERSION}/deploy/kubernetes/snapshot-controller/setup-snapshot-controller.yaml"
+  cnt=0
+  until kubectl get pods | grep '^snapshot-controller.* Running '; do
+    if [ $cnt -gt 30 ]; then
+        echo "Running statefulsets:"
+        kubectl describe statefulsets
```
msau42: this also won't work if we turn it into a deployment

release-tools/prow.sh
```diff
@@ -657,6 +660,59 @@ install_hostpath () {
...
+    fi
+    echo "$(date +%H:%M:%S)" "waiting for snapshot RBAC setup complete, attempt #$cnt"
+	cnt=$((cnt + 1))
+    sleep 10   
+  done
+
+
+  kubectl apply -f "https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/${CSI_SNAPSHOTTER_VERSION}/deploy/kubernetes/snapshot-controller/setup-snapshot-controller.yaml"
+  cnt=0
+  until kubectl get pods | grep '^snapshot-controller.* Running '; do
```
msau42: Can we use labels instead?

msau42: This means that once we have at least one that's running, then we succeed? should we wait for all replicas?

deploy/util/deploy-hostpath.sh
ggriffiths: Tested this and they both work as expected:
```
[grant@PDC4-SM26-N8 csi-driver-host-path]$ version_gt 1.2 1.2.2
false
[grant@PDC4-SM26-N8 csi-driver-host-path]$ version_gt 1.2.0 1.2.2
false
```

ggriffiths: The only problem I found here is the following. 1.2.0 is sorted greater than 1.2
```
[grant@PDC4-SM26-N8 csi-driver-host-path]$ version_gt 1.2.0 1.2
true
```

deploy/util/deploy-hostpath.sh
ggriffiths: copy/paste error - fixed

deploy/util/deploy-hostpath.sh
ggriffiths: corrected this - comment was incorrect.

release-tools/prow.sh
ggriffiths: Changed to `kubectl describe pods -l app=snapshot-controller`

release-tools/prow.sh
ggriffiths: Changed to `while [ $(kubectl get pods -l app=snapshot-controller | grep 'Running' | wc -l) -lt $expected_running_pods ]; do`

deploy/util/deploy-hostpath.sh
ggriffiths: This was leftover from when we weren't deploying the snapshotclass for k8s 1.17. Just removed this if/else block altogether.

ggriffiths:
Addressed most recent round of feedback @xing-yang @msau42 

/test pull-kubernetes-csi-csi-driver-host-path-1-17-on-kubernetes-master

release-tools/prow.sh
```diff
@@ -657,6 +660,60 @@ install_hostpath () {
...
+    fi
+    echo "$(date +%H:%M:%S)" "waiting for snapshot RBAC setup complete, attempt #$cnt"
+	cnt=$((cnt + 1))
+    sleep 10   
+  done
+
+
+  kubectl apply -f "https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/${CSI_SNAPSHOTTER_VERSION}/deploy/kubernetes/snapshot-controller/setup-snapshot-controller.yaml"
+  cnt=0
+  expected_running_pods=3
```
msau42: Can we pull the "replicas" value from the deployment file? (It's currently one)

ggriffiths: Updated this.
/test pull-kubernetes-csi-csi-driver-host-path-1-17-on-kubernetes-master

ggriffiths:
Was missing the snapshotclass file in the k8s 1.17 directory. Just re-added it.
/test pull-kubernetes-csi-csi-driver-host-path-1-17-on-kubernetes-master

ggriffiths:
Was missing the snapshotclass file in the k8s 1.17 directory. Just re-added it.
/test pull-kubernetes-csi-csi-driver-host-path-1-17-on-kubernetes-master

msau42:
lgtm! thanks for all your work on this!
/approve

release-tools/prow.sh
```diff
@@ -322,6 +322,9 @@ configvar CSI_PROW_E2E_ALPHA_GATES_1_16 'VolumeSnapshotDataSource=true' "alpha f
 configvar CSI_PROW_E2E_ALPHA_GATES_LATEST 'VolumeSnapshotDataSource=true' "alpha feature gates for latest Kubernetes"
 configvar CSI_PROW_E2E_ALPHA_GATES "$(get_versioned_variable CSI_PROW_E2E_ALPHA_GATES "${csi_prow_kubernetes_version_suffix}")" "alpha E2E feature gates"
 
+# Which external-snapshotter tag to use for the snapshotter CRD and snapshot-controller deployment
+configvar CSI_SNAPSHOTTER_VERSION 'v2.0.0-rc4' "external-snapshotter version tag"
```
xing-yang: For changes under release-tools, I believe you need to make the change in kubernetes-csi/csi-release-tools first and port it to here.

ggriffiths: > lgtm! thanks for all your work on this!
> /approve

Thanks! I'll update my PR to the release tools repo with the latest prow.sh changes. Once that's merged, I'll update the PR with the final commits.

ggriffiths:
> lgtm! thanks for all your work on this!
> /approve

Thanks! I'll update my PR to the release tools repo with the latest prow.sh changes. Once that's merged, I'll update the PR with the final commits.

release-tools/prow.sh
ggriffiths: :+1: Will update my release-tools PR with these changes.

ggriffiths:
Here is the PR for these prow.sh changes. Updated it with my latest changes to release-tools/prow.sh
https://github.com/kubernetes-csi/csi-release-tools/pull/49
@xing-yang @msau42

ggriffiths:
Tests seem to failing after I did 
`git subtree pull --prefix=release-tools https://github.com/kubernetes-csi/csi-release-tools.git master`

```
if [ "$ARCH" = "amd64" ]; then \
	CGO_ENABLED=0 GOOS=windows go build -mod=vendor -a -ldflags '-X main.version=v1.3.0-rc1-40-ga068236c -extldflags "-static"' -o ./bin/hostpathplugin.exe ./cmd/hostpathplugin ; \
	CGO_ENABLED=0 GOOS=linux GOARCH=ppc64le go build -mod=vendor -a -ldflags '-X main.version=v1.3.0-rc1-40-ga068236c -extldflags "-static"' -o ./bin/hostpathplugin-ppc64le ./cmd/hostpathplugin
/bin/sh: 3: Syntax error: end of file unexpected (expecting "fi")
make: *** [release-tools/build.make:71: build-hostpathplugin] Error 2
```

msau42:
My bad, I think it's missing a "\" at the end

ggriffiths:
Updated with the fix to build.make
/test pull-kubernetes-csi-csi-driver-host-path-1-17-on-kubernetes-master

ggriffiths:
Seems like a change is incompatible with `pull-kubernetes-csi-csi-driver-host-path-1-17-on-kubernetes-master`.

I just re-ran the following on a local k8s 1.16 cluster
`install_snapshot_crds`
`install_snapshot_controller`
`deploy/kubernetes-1.17/deploy-hostpath.sh`
and tested snapshots manually, and everything seems fine.

@msau42 do you know of any other prow.sh or test-infra changes that may cause this test to fail?
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_csi-driver-host-path/98/pull-kubernetes-csi-csi-driver-host-path-1-17-on-kubernetes-master/1202366525122547714

msau42:
```
Dec  4 23:41:34.186: INFO: At 2019-12-04 23:36:35 +0000 UTC - event for pod-subpath-test-hostpath-csi-k8s-io-dynamicpv-7ntl: {kubelet csi-prow-worker} FailedMount: MountVolume.MountDevice failed for volume "pvc-0200512e-849b-4df5-b90e-547ba8e30c64" : kubernetes.io/csi: attacher.MountDevice failed to create newCsiDriverClient: driver name hostpath.csi.k8s.io not found in the list of registered CSI drivers
```

Looks like the volume got provisioned on node csi-prow-worker2, but the Pod got scheduled on csi-prow-worker:
```
I1204 23:36:19.514673       1 event.go:258] Event(v1.ObjectReference{Kind:"PersistentVolumeClaim", Namespace:"provisioning-2145", Name:"hostpath.csi.k8s.iov56w5", UID:"0200512e-849b-4df5-b90e-547ba8e30c64", APIVersion:"v1", ResourceVersion:"2248", FieldPath:""}): type: 'Normal' reason: 'Provisioning' External provisioner is provisioning volume for claim "provisioning-2145/hostpath.csi.k8s.iov56w5"
I1204 23:36:19.522595       1 connection.go:183] GRPC response: {"volume":{"accessible_topology":[{"segments":{"topology.hostpath.csi/node":"csi-prow-worker2"}}],"capacity_bytes":1048576,"volume_id":"df74845e-16ee-11ea-8c4c-3edb8610010e"}}
```

deploy/kubernetes-1.17/hostpath/csi-hostpath-provisioner.yaml
```diff
@@ -0,0 +1,54 @@
...
+                values:
+                - csi-hostpathplugin
+            topologyKey: kubernetes.io/hostname
+      serviceAccountName: csi-provisioner
+      containers:
+        - name: csi-provisioner
+          image: quay.io/k8scsi/csi-provisioner:v1.5.0-rc1
+          args:
+            - -v=5
+            - --csi-address=/csi/csi.sock
```
msau42: Ah you're missing the "--feature-gates=Topology=true" flag

ggriffiths: Nice catch :+1: Will add now.

ggriffiths:
Looks like that did the job! Thanks @msau42 !

All tests are passing now.
/lgtm

PR #102, opened 2019-10-15
Release 1.2.0

pohly:
**What type of PR is this?**
/kind feature

This in preparation for the v1.2.0 release.

```release-note
updated sidecars to latest stable releases
```


deploy/kubernetes-1.15/hostpath/csi-hostpath-provisioner.yaml
```diff
@@ -40,7 +40,7 @@ spec:
       serviceAccountName: csi-provisioner
       containers:
         - name: csi-provisioner
-          image: quay.io/k8scsi/csi-provisioner:v1.3.0
+          image: quay.io/k8scsi/csi-provisioner:v1.4.0
```
pohly: I am unsure which deployments should be using 1.4.0. It's marked as compatible with Kubernetes >= 1.13 (https://github.com/kubernetes-csi/external-provisioner/releases/tag/v1.4.0). Do we still need any of the older releases? What for?

CHANGELOG-2.0.md
```diff
@@ -0,0 +1,21 @@
...
+## Bug Fixes
+
+- /csi-data-dir optional ([#73](https://github.com/kubernetes-csi/csi-driver-host-path/pull/73), [@msau42](https://github.com/msau42))
+- Set volume content source if creating volume from snapshot ([#51](https://github.com/kubernetes-csi/csi-driver-host-path/pull/51), [@zhucan](https://github.com/zhucan)).
+- Fixes cp expansion issue for volume cloning ([#82](https://github.com/kubernetes-csi/csi-driver-host-path/pull/82), [@j-griffith](https://github.com/j-griffith)).
+
+
+## Other Notable Changes
+
+- The deployment uses `hostpath.csi.k8s.io` as driver name ([#64](https://github.com/kubernetes-csi/csi-driver-host-path/pull/64), [@pohly](https://github.com/pohly)).
```
msau42: Is this technically a breaking change? existing PVs using the old driver name won't work anymore?

deploy/kubernetes-1.13/hostpath/csi-hostpath-plugin.yaml
```diff
@@ -37,7 +37,7 @@ spec:
       hostNetwork: true
       containers:
         - name: node-driver-registrar
-          image: quay.io/k8scsi/csi-node-driver-registrar:v1.1.0
+          image: quay.io/k8scsi/csi-node-driver-registrar:v1.2.0
```
msau42: Shall we remove the entire 1.13 deployment? 1.13 is out of K8s support policy technically and we don't run tests on 1.13 anymore

deploy/kubernetes-1.15/hostpath/csi-hostpath-provisioner.yaml
msau42: It could technically use 1.4.0.  The main argument I can see for keeping older versions of sidecars is so that they still get CI testing.

deploy/kubernetes-1.13/hostpath/csi-hostpath-plugin.yaml
pohly: Fine with me.

deploy/kubernetes-1.15/hostpath/csi-hostpath-provisioner.yaml
pohly: That depends on whether anyone really needs to use those older versions. If that is not the case, then we don't need to support and test anything besides 1.4.0.

See also my questions in https://github.com/kubernetes-csi/docs/pull/212.

pohly: BTW, the inverse is also true: because we claim that external-provisioner is compatible with Kubernetes >= 1.13, we need to test that, in particular because it is a combination that hasn't been tested before.

IMHO the "kubernetes-1.14" deployment should cover the recommended set of sidecars for that Kubernetes release. I'm just not sure at the moment what the recommendation is :man_shrugging: 

If we want to support also older releases of the sidecars, then we need additional deployments and Prow jobs.

msau42: The recommended should be documented in the sidecar pages. Let's stick with that for now.  

Agree that we have a testing gap with what we claim to support. We can brainstorm ideas on how to improve the coverage in a separate issue for now.

pohly: > The recommended should be documented in the sidecar pages.

You mean the "recommended K8s version" column? That's not quite the same.

For example, external-provisioner 1.4.0 may have some features that depend on 1.16 (hypothetically speaking, I have no idea whether that is the case). According to https://kubernetes-csi.github.io/docs/kubernetes-compatibility.html that means it's "recommended K8s version" will list 1.16 because "recommended Kubernetes version specifies the lowest Kubernetes version needed where all features of a sidecar will function correctly". That column basically answers the question "I want external-provisioner 1.4.0, which Kubernetes version will work best".

The other question is "I have Kubernetes 1.15, which sidecar versions should I use for that".

For example, suppose 1.4.0 can fully replace 1.3.1 on Kubernetes 1.15. Does that make it the recommended release for 1.15? This is not documented. The closest we get to documenting that are the reference deployments in this repo.

msau42: I think we should make the answers to both questions the same for simplicity

pohly: As discussed in the CI standup meeting, the deployments should use the sidecar versions on those (and only) those Kubernetes releases which are "recommended", i.e. we don't attempt to replace older sidecar versions if there have been API changes even if that would technically work - we simply don't test it and thus don't support it. Instead, older sidecar releases get maintained for older Kubernetes releases.

https://kubernetes-csi.github.io/docs/livenessprobe.html doesn't list "recommended Kubernetes". I made the assumption that there have been no API changes and thus all Kubernetes releases should be using the latest (1.1.0).

Same with https://kubernetes-csi.github.io/docs/node-driver-registrar.html (latest v1.2.0).

https://kubernetes-csi.github.io/docs/external-snapshotter.html lists 1.14 for 1.2.2, so it gets used on 1.14, 1.15 and 1.16.

CHANGELOG-2.0.md
```diff
@@ -0,0 +1,23 @@
+# Changelog since v1.1.0
+
+## Breaking Changes
+
+- The deployment uses `hostpath.csi.k8s.io` as driver name ([#64](https://github.com/kubernetes-csi/csi-driver-host-path/pull/64), [@pohly](https://github.com/pohly)).
```
msau42: Can you add a note about what might break and what they can do to mitigate/resolve it?

deploy/kubernetes-1.13/hostpath/csi-hostpath-plugin.yaml
msau42: do you plan to remove 1.13 specs in this pr?

CHANGELOG-2.0.md
pohly: Done.

pohly:
> do you plan to remove 1.13 specs in this pr?

Done, also mentioned in changelog.

CHANGELOG-1.2.md
```diff
@@ -0,0 +1,27 @@
+# Changelog since v1.1.0
+
+## Breaking Changes
+
+- The deployment uses `hostpath.csi.k8s.io` as driver name ([#64](https://github.com/kubernetes-csi/csi-driver-host-path/pull/64), [@pohly](https://github.com/pohly)).
+  Make sure that there are no active ephemeral inline volumes using the old `csi-hostpath` name before updating because otherwise
```
msau42: Also existing PVs created using the old driver name won't work either.

pohly: Aren't they tied to the storage class, and if the storage class gets updated the driver is still found under the new name? That at least was my hope; I haven't tried any of this.

msau42: One provisioned, the PV object contains the name of the driver in it.  So if you change the driver name, it will no longer be able to process existing PVs

pohly: Updated.

CHANGELOG-1.2.md
```diff
@@ -0,0 +1,27 @@
+# Changelog since v1.1.0
+
+## Breaking Changes
+
+- The deployment uses `hostpath.csi.k8s.io` as driver name ([#64](https://github.com/kubernetes-csi/csi-driver-host-path/pull/64), [@pohly](https://github.com/pohly)).
+  Make sure that there are no persistent or ephemeral volumes using the old `csi-hostpath` name before updating because otherwise
+  those volumes cannot be removed. Pods with ephemeral volumes will be stuck in "terminating" state.
```
msau42: And any new pods cannot start

pohly: "Any new pod" or "pods using the old name"? I suppose the later.

pohly: Updated.

CHANGELOG-1.2.md
```diff
@@ -0,0 +1,28 @@
+# Changelog since v1.1.0
+
+## Breaking Changes
+
+- The deployment uses `hostpath.csi.k8s.io` as driver name ([#64](https://github.com/kubernetes-csi/csi-driver-host-path/pull/64), [@pohly](https://github.com/pohly)).
+  Make sure that there are no persistent or ephemeral volumes using the old `csi-hostpath` name before updating because otherwise
+  those volumes cannot be removed. Pods with such ephemeral volumes will be stuck in "terminating" state. New pods using
+  the old name in an inline volume will not be able to start.
```
msau42: This is true for pvs too

pohly: Updated.

I hereby promise to never, ever change the name of a CSI driver again. :grimacing:

deploy/kubernetes-1.15/hostpath/csi-hostpath-attacher.yaml
```diff
@@ -40,7 +40,7 @@ spec:
       serviceAccountName: csi-attacher
       containers:
         - name: csi-attacher
-          image: quay.io/k8scsi/csi-attacher:v1.2.0
+          image: quay.io/k8scsi/csi-attacher:v2.0.0
```
msau42: Hm, one reason though to update 1.14 deployment to use attacher 2.0.0 is that currently our canary 1.14 tests are failing because 2.0 had added new rbac rules. We would need to update the 1.14 deployment here to pick up the new rbacs.

pohly: > currently our canary 1.14 tests are failing because 2.0 had added new rbac rules

But isn't that as it should be? attacher 2.0.0 is not meant to run on Kubernetes 1.14, so we shouldn't be testing that combination (anymore).

msau42: This is where it gets complicated.  I think it's still valuable to test 1.14 canary for all the other sidecars, but not attacher. Or we make an exception in this case

pohly:
> This is where it gets complicated.  I think it's still valuable to
> test 1.14 canary for all the other sidecars, but not attacher. Or we
> make an exception in this case

When it comes to testing, we can always introduce an alternative
deployment and use that for canary testing or tweak the test setup
script. Which Prow job exactly are we talking about?

msau42:
https://k8s-testgrid.appspot.com/sig-storage-csi-ci#canary-on-1.14

Issue #105, opened 2019-10-23
Idempotency: repeated Node Unpublish operations cause failure


okartau:
Background: in [pmem-csi driver](https://github.com/intel/pmem-csi) we added more idempotency conformance. 
To test, the repeated Node Volume operations were [added in csi-test](https://github.com/kubernetes-csi/csi-test/pull/229).
In process of upstreaming this change, the CI tests that use host-path-driver, fail:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_csi-test/229/pull-kubernetes-csi-csi-test/1185891170576764931

host-path driver currently does not pay attention to repeated requests possibility, there are no mutexes that protect against related race conditions.

pohly:
We discussed this today in the Kubernetes-CSI standup and everyone agreed that a) this should be fixed in the hostpath driver and b) that having these idempotency tests in csi-test will be very useful.

BenTheElder:
what is necessary to fix this? just limiting the concurrency of operations in the driver?

BenTheElder:
xref: https://github.com/kubernetes-csi/csi-driver-host-path/issues/72, this sounds like a superset of that

okartau:
I created #139 to tolerate repeated NodeUnpublish requests.
My local testing shows that this allows hostpath driver to pass the added repeated csi-sanity testing currently coming in https://github.com/kubernetes-csi/csi-test/pull/229, so right now there is no need to add mutex locking to satisfy those repeated tests, as I can't see failures.

pohly:
> so right now there is no need to add mutex locking to satisfy those repeated tests, as I can't see failures.

This is a dangerous argument: that you haven't been able to trigger a problem caused by missing locking does not mean that the problem doesn't exist or that we don't need to do anything about it.

In practice the problem should be exceedingly rare. It only happens when two gRPC calls reach the driver at the same time and get processed in parallel. I think the idempotency tests don't cover that, they just issue the same call multiple times sequentially.

So while we don't need mutex locking to make the NodeUnpublishVolume idempotent, we still need it to make the code more robust.

okartau:
> This is a dangerous argument: that you haven't been able to trigger a problem caused by missing locking does not mean that the problem doesn't exist or that we don't need to do anything about it.

I agree, I did not mean we don't need to deal with locking at all, just that currently to-be-added test cases do not expose problem that we can see fixed by added locking.
[...]

PR #112, opened 2019-11-01
add mutex locks to protect hostPathVolumes and hostPathSnapshots

### LINKED ISSUE #72: map access needs lock
All `hostPathVolumes` and `hostPathSnapshots` map updates/reads need a lock.

/help
/good-first-issue

A-Hilaly:
**What type of PR is this?**

/kind feature

Use mutex locks to prevent from data races when concurrent updates/reads hit `hostPathVolumes` and `hostPathSnapshots`

Fixes #72 

```release-note
NONE
```

### DIFF
--- a/pkg/hostpath/hostpath.go
+++ b/pkg/hostpath/hostpath.go
@@ -21,6 +21,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"sync"
 
 	"github.com/golang/glog"
 	"google.golang.org/grpc/codes"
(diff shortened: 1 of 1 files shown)


pohly:
This might be not enough by far to ensure that concurrent operations are handled safely by the driver. It locks the low-level data structures, but doesn't prevent the execution of a resent `CreateVolume` request, for example.

@okartau wanted to work on a full solution, see https://github.com/kubernetes-csi/csi-driver-host-path/issues/105

A-Hilaly:
Thanks for the clarification @pohly, I wasn't aware that #105 is related to this one.

PR #114, opened 2019-11-04
updated deployment: fix pod count, canary jobs

pohly:
**What type of PR is this?**
/kind failing-test

Canary jobs are failing without it: https://storage.googleapis.com/kubernetes-jenkins/logs/ci-kubernetes-csi-canary-on-kubernetes-1-14/1191339260892942337/build-log.txt

Also includes some other enhancements, see individual commits for details.

```release-note
NONE
```

### DIFF
--- a/deploy/util/deploy-hostpath.sh
+++ b/deploy/util/deploy-hostpath.sh
@@ -32,14 +32,15 @@ BASE_DIR=$(dirname "$0")
 # - IMAGE_TAG
 # These are used as fallback when the more specific variables are unset or empty.
 #
-# IMAGE_TAG=canary is ignored for images that are blacklisted in the deployment's
-# top-level README.md. This is meant for images which have known API breakages and
-# thus cannot work in those deployments anymore. For the purpose of this script,
-# any line matching " *-* <image name>< optional additional text>" is fine, for example:
+# IMAGE_TAG=canary is ignored for images that are blacklisted in the
+# deployment's optional canary-blacklist.txt file. This is meant for
+# images which have known API breakages and thus cannot work in those
+# deployments anymore. That text file must have the name of the blacklisted
+# image on a line by itself, other lines are ignored. Example:
 #
-#     The following canary images are known to be incompatible with this
-#     deployment:
-#     - csi-snapshotter (canary uses VolumeSnapshot v1beta)
+#     # The following canary images are known to be incompatible with this
+#     # deployment:
+#     csi-snapshotter
 #
 # Beware that the .yaml files do not have "imagePullPolicy: Always". That means that
 # also the "canary" images will only be pulled once. This is good for testing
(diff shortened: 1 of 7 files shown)


pohly:
/assign @xing-yang @ggriffiths 

@msau42 said that you were working on this, too?

xing-yang:
@pohly, Grant and I are trying to add a new folder for 1.17.  The PR is here:
https://github.com/kubernetes-csi/csi-driver-host-path/pull/98

Your PR is still needed to fix problems in existing folders.

deploy/util/deploy-hostpath.sh
```diff
@@ -132,7 +141,15 @@ for i in $(ls ${BASE_DIR}/hostpath/*.yaml | sort); do
             # Only do this for the images which are meant to be configurable.
             if update_image "$name"; then
                 prefix=$(eval echo \${${varname}_REGISTRY:-${IMAGE_REGISTRY:-${registry}}}/ | sed -e 's;none/;;')
-                suffix=$(eval echo :\${${varname}_TAG:-${IMAGE_TAG:-${tag}}})
+                if [ "$IMAGE_TAG" = "canary" ] &&
+                   [ -f ${BASE_DIR}/README.md ] &&
+                   grep -q "^ *-* *$name *" ${BASE_DIR}/README.md; then
```
msau42: Grepping the readme is a bit brittle... I would prefer some explicit ENV/flag/config file.

deploy/util/deploy-hostpath.sh
```diff
@@ -150,11 +167,13 @@ done
 # about the deployment here, otherwise we wouldn't know what to wait
 # for: the expectation is that we run attacher, provisioner,
 # snapshotter, resizer, socat and hostpath plugin in the default namespace.
+expected_running_pods=6
 cnt=0
-while [ $(kubectl get pods 2>/dev/null | grep '^csi-hostpath.* Running ' | wc -l) -lt 5 ] || ! kubectl describe volumesnapshotclasses.snapshot.storage.k8s.io 2>/dev/null >/dev/null; do
+while [ $(kubectl get pods 2>/dev/null | grep '^csi-hostpath.* Running ' | wc -l) -lt $expected_running_pods ] || ! kubectl describe volumesnapshotclasses.snapshot.storage.k8s.io 2>/dev/null >/dev/null; do
```
msau42: I think we should remove the check for volumesnapshot crds existing to be more compatible with the beta model where the it's the cluster deployer that's responsible for installing crds, not the csi driver.  cc @xing-yang

msau42:
Were you planning to address csi-attacher incompatibility issues in this PR too?

deploy/util/deploy-hostpath.sh
pohly: I chose the README.md because it is some useful additional information. Having said that, I am fine with some explicit config file in the deploy/kubernetes-x.yy directories. It has to be in a format that can be parsed from this shell script, so in the end it's still going to be using grep, though.

deploy/util/deploy-hostpath.sh
pohly: Makes sense, but that then needs to go into https://github.com/kubernetes-csi/csi-driver-host-path/pull/98.

deploy/util/deploy-hostpath.sh
pohly: Done.

msau42:
> > Were you planning to address csi-attacher incompatibility issues in this PR too?
> 
> Which incompatibility? Is canary of external-attacher also incompatible with all three old deployments? I can certainly also add that to the blacklists.

Yes that's why we bumped external-attacher version to 2.0.0.  It added new required RBAC rules.  Alternatively, we could bump the external-attacher version in older deployments to 2.0.0 if we wanted to because K8s API wise, it can still run against 1.14+.

deploy/util/deploy-hostpath.sh
msau42: That's fine, I'm just looking for a file with a well-known syntax to parse, rather than a free-form readme.

msau42:
/lgtm
/approve
to fix snapshotter version on older deployments.  Canary jobs will probably still fail due to attacher, but we can resolve that separately.

pohly:
> Yes that's why we bumped external-attacher version to 2.0.0. It added new required RBAC rules.
> Alternatively, we could bump the external-attacher version in older deployments to 2.0.0 if we wanted to because K8s API wise, it can still run against 1.14+.

As discussed elsewhere, we recommend v2.0.0 only for Kubernetes >= 1.15, so we shouldn't bother testing it on 1.14. Therefore I prefer to blacklist it there -> https://github.com/kubernetes-csi/csi-driver-host-path/pull/116

PR #117, opened 2019-11-05
system pod logging

pohly:
**What type of PR is this?**
/kind failing-test

This imports the updated csi-release-tools with logging of system containers, which will be useful for the periodic Prow jobs (now) and pull request testing (after doing a release and bumping up the version of csi-driver-host-path that is used in those jobs).

```release-note
NONE
```

### DIFF
--- a/release-tools/prow.sh
+++ b/release-tools/prow.sh
@@ -683,16 +683,16 @@ EOF
 
 }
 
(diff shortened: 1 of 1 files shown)


pohly:
/assign @msau42 

I considered just closing this PR and then import via another round of automated csi-release-tools updates, but this should be useful already now, so I propose to merge it.

PR #120, opened 2019-11-14
[Test] delete kind cluster after tests run

mucahitkurt:
Delete kind cluster after tests run. Inside a real Prow job it is better to clean up at runtime instead of leaving that to the Prow job cleanup code because the later sometimes times out.

Related issue is https://github.com/kubernetes-csi/csi-release-tools/issues/24

### DIFF
--- a/release-tools/prow.sh
+++ b/release-tools/prow.sh
@@ -579,7 +579,18 @@ EOF
     KUBECONFIG="$(kind get kubeconfig-path --name=csi-prow)"
     export KUBECONFIG
 }
-
+# Deletes kind cluster inside a prow job
+delete_cluster_inside_prow_job() {
+    # Inside a real Prow job it is better to clean up at runtime
+    # instead of leaving that to the Prow job cleanup code
+    # because the later sometimes times out (https://github.com/kubernetes-csi/csi-release-tools/issues/24#issuecomment-554765872).
+    if [ "$JOB_NAME" ]; then
+        if kind get clusters | grep -q csi-prow; then
+            run kind delete cluster --name=csi-prow || die "kind delete failed"
+        fi
+        unset KUBECONFIG
+    fi
+}
 # Looks for the deployment as specified by CSI_PROW_DEPLOYMENT and CSI_PROW_KUBERNETES_VERSION
 # in the given directory.
 find_deployment () {
@@ -1017,6 +1028,7 @@ main () {
                     fi
                 fi
             fi
+            delete_cluster_inside_prow_job
         fi
 
         if tests_need_alpha_cluster && [ "${CSI_PROW_E2E_ALPHA_GATES}" ]; then
@@ -1047,6 +1059,7 @@ main () {
                     fi
                 fi
             fi
+            delete_cluster_inside_prow_job
         fi
     fi
 


pohly:
Can you explain why you want this changed?

The prow.sh script intentionally doesn't clean up after itself. When it's used inside a Prow job, deleting that job should clean up. When it's used on a developer's machine, then keeping everything running simplifies debugging in case of a failure. Running the script again will then destroy the cluster and start anew, so it's not leaving an unbounded number of clusters behind.

I had once contemplated to write a wrapper script around prow.sh which takes care of setting up a temporary GOPATH (which does get modified by the script!) and properly cleans up afterwards. That makes sense to me, but not trying to clean up inside the script itself (because it isn't always needed or desired).

mucahitkurt:
> Can you explain why you want this changed?
> 
> The prow.sh script intentionally doesn't clean up after itself. When it's used inside a Prow job, deleting that job should clean up. When it's used on a developer's machine, then keeping everything running simplifies debugging in case of a failure. Running the script again will then destroy the cluster and start anew, so it's not leaving an unbounded number of clusters behind.
> 
> I had once contemplated to write a wrapper script around prow.sh which takes care of setting up a temporary GOPATH (which does get modified by the script!) and properly cleans up afterwards. That makes sense to me, but not trying to clean up inside the script itself (because it isn't always needed or desired).

@pohly I opened the PR because of the issue https://github.com/kubernetes-csi/csi-release-tools/issues/24. I got the issue as not cleaning cluster after test run can cause timeout for the test jobs.

Btw this is the test PR, actual PR is https://github.com/kubernetes-csi/csi-release-tools/pull/46.

pohly:
https://github.com/kubernetes-csi/csi-release-tools/issues/24#issuecomment-554765872 explains why this change is needed. This is information that better should better be copied into the commit message and the PR description because then it is available when reviewers like me need it. I wasn't aware of that issue.

For csi-release-tools changes it is better to first discuss and finalize the review in one of the components, then create the PR in csi-release-tools. That way there's no confusion about where to discuss the change and if changes are needed, you don't need to update the csi-release-tools PR potentially multiple times.

release-tools/prow.sh
```diff
@@ -1017,6 +1023,7 @@ main () {
                     fi
                 fi
             fi
+            delete_cluster
```
pohly: Can we make this conditional and only do it in the CI? At least I sometimes run this script manually and prefer to keep the cluster around for debugging.

We could turn this into:
```
# Inside a real Prow job it is better to clean up at runtime
# instead of leaving that to the Prow job cleanup code
# because the later sometimes times out (https://github.com/kubernetes-csi/csi-release-tools/issues/24#issuecomment-554765872).
if [ "$JOB_NAME" ]; then
    delete_cluster
fi
```

mucahitkurt:
> [kubernetes-csi/csi-release-tools#24 (comment)](https://github.com/kubernetes-csi/csi-release-tools/issues/24#issuecomment-554765872) explains why this change is needed. This is information that better should better be copied into the commit message and the PR description because then it is available when reviewers like me need it. I wasn't aware of that issue.
> 
> For csi-release-tools changes it is better to first discuss and finalize the review in one of the components, then create the PR in csi-release-tools. That way there's no confusion about where to discuss the change and if changes are needed, you don't need to update the csi-release-tools PR potentially multiple times.

Ok. Thanks for the heads up!

release-tools/prow.sh
mucahitkurt: Fixed! I assume you don't want to same thing for `tests_need_alpha_cluster`?

pohly: Sorry, I do. I just didn't point it out explicitly.

mucahitkurt: I moved the control to delete_cluster method and renamed it to `delete_cluster_inside_prow_job`.

mucahitkurt:
> Looks good to me, please update the `csi-release-tools` PR and we can merge it there.
> 
> I can then kick of a round of repo updates to merge that into the different sidecars.

@pohly I updated the [`csi-release-tools` PR](https://github.com/kubernetes-csi/csi-release-tools/pull/46), thank you!

PR #123, opened 2019-12-03
Run all containers as privileged

jsafrane:
**What type of PR is this?**
/kind flake
(All hostpath tests are broken on RHEL / CentOS with SELinux)

On systems with SELinux enabled, non-privileged containers can't access data of privileged containers. Since the socket is exposed by privileged container, all sidecars must be privileged too.

And I removed `hostNetwork: true`  while at it - was it used for anything? It exposes liveness probe port on the host and therefore one node can run only one driver, limiting parallelization of the tests.

@pohly @msau42, PTAL. I know it might be controversial, but it's necessary for RHEL / CentOS.

```release-note
NONE
```

### DIFF
--- a/deploy/kubernetes-1.16/hostpath/csi-hostpath-provisioner.yaml
+++ b/deploy/kubernetes-1.16/hostpath/csi-hostpath-provisioner.yaml
@@ -46,6 +46,11 @@ spec:
             - --csi-address=/csi/csi.sock
             - --connection-timeout=15s
             - --feature-gates=Topology=true
+          securityContext:
+            # This is necessary only for systems with SELinux, where
+            # non-privileged sidecar containers cannot access unix domain socket
+            # created by privileged CSI driver container.
+            privileged: true
           volumeMounts:
             - mountPath: /csi
               name: socket-dir
--- a/deploy/kubernetes-1.16/hostpath/csi-hostpath-attacher.yaml
+++ b/deploy/kubernetes-1.16/hostpath/csi-hostpath-attacher.yaml
@@ -44,6 +44,11 @@ spec:
           args:
             - --v=5
             - --csi-address=/csi/csi.sock
+          securityContext:
+            # This is necessary only for systems with SELinux, where
+            # non-privileged sidecar containers cannot access unix domain socket
+            # created by privileged CSI driver container.
+            privileged: true
           volumeMounts:
           - mountPath: /csi
             name: socket-dir
--- a/deploy/kubernetes-1.16/hostpath/csi-hostpath-plugin.yaml
+++ b/deploy/kubernetes-1.16/hostpath/csi-hostpath-plugin.yaml
@@ -46,6 +46,9 @@ spec:
             - --csi-address=/csi/csi.sock
             - --kubelet-registration-path=/var/lib/kubelet/plugins/csi-hostpath/csi.sock
           securityContext:
+            # This is necessary only for systems with SELinux, where
+            # non-privileged sidecar containers cannot access unix domain socket
+            # created by privileged CSI driver container.
             privileged: true
           env:
             - name: KUBE_NODE_NAME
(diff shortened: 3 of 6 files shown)


deploy/kubernetes-1.16/hostpath/csi-hostpath-provisioner.yaml
```diff
@@ -46,6 +46,8 @@ spec:
             - --csi-address=/csi/csi.sock
             - --connection-timeout=15s
             - --feature-gates=Topology=true
+          securityContext:
+            privileged: true
```
msau42: Are you running intree hostpath tests on rhel? The intree specs are unprivileged: https://github.com/kubernetes/kubernetes/blob/master/test/e2e/testing-manifests/storage-csi/hostpath/hostpath/csi-hostpath-provisioner.yaml

jsafrane: This PR is about CSI sidecars accessing CSI driver socket - unprivileged sidecar cannot open socket of privileged driver container. 

The SELinux setting of the actual "volume" provided by the driver to pods is orthogonal to this PR and currently I don't see any issues there. The behavior is the same as with in-tree hostpath plugin.

msau42: I mean how are the intree hostpath driver tests passing when the sidecars are not privileged?  What is different between these out of tree manifests vs the intree manifests?

jsafrane: Ah, they're not. I'm fixing the source (=this repo) first and then I'll pick it into kubernetes/kubernetes test manifests.

msau42: Does it help if we put all the sidecars and driver together in a single pod? 

I am mainly concerned because this driver is supposed to be used as an example and people will blindly copy paste. Can you at least add a comment explaining why this is necessary?

jsafrane: > Does it help if we put all the sidecars and driver together in a single pod?

No :-(

I added comments to all `privileged: true`

msau42: Why not? Then they can share the CSI socket through an emptydir instead of hostpath

jsafrane: In both cases it ends up as a socket somewhere on the host. And unprivileged container cannot access files of a privileged one, regardless if they are in the same pod or not. The container runtime does not care.

msau42: What if you have separate deployments for controller vs node? Does the controller plugin need to be privileged?

jsafrane: In that case it's OK. I am not sure hostpath driver supports such deployment, need to check the code.

msau42: The only thing I can think of that may need some additional capabilities is provisioning a block device.  Provisioning a directory only needs proper host permissions, not privileged?

jsafrane: There is list of provisioned volumes in the hostpath driver and running two pods with separate driver containers is not possible:

Provisioning:
https://github.com/kubernetes-csi/csi-driver-host-path/blob/3b4d37f476f31cfd134c2abb9f5f3a7021fe9291/pkg/hostpath/hostpath.go#L210

Mount:
https://github.com/kubernetes-csi/csi-driver-host-path/blob/3b4d37f476f31cfd134c2abb9f5f3a7021fe9291/pkg/hostpath/hostpath.go#L138-L143

msau42: Ok. I wish we can improve hostpath driver to not require this in memory structure (means that it can't handle restarts?), but that's a separate issue.

PR #124, opened 2019-12-05
Run all 1.17 containers as privileged

ggriffiths:
**What type of PR is this?**
/kind cleanup

This PR makes the 1.17 containers run as privileged

Fixes #

```release-note
NONE
```


deploy/kubernetes-1.17/hostpath/csi-hostpath-plugin.yaml
```diff
@@ -34,7 +34,6 @@ spec:
       labels:
         app: csi-hostpathplugin
     spec:
-      hostNetwork: true
```
ggriffiths: Also removed this line, as I saw this change in 
https://github.com/kubernetes-csi/csi-driver-host-path/pull/123/files

PR #126, opened 2019-12-06
Create snapshotclass based on snapshotter sidecar version

ggriffiths:
Signed-off-by: Grant Griffiths <grant@portworx.com>

**What type of PR is this?**
/kind failing-test

We were deploying the latest snapshotclass for `on-kubernetes-master` tests, regardless of the csi-snapshotter version. Now we'll apply the snapshotclass depending on the sidecar version.

Fixes #

```release-note
NONE
```

### DIFF
--- a/deploy/util/deploy-hostpath.sh
+++ b/deploy/util/deploy-hostpath.sh
@@ -104,6 +104,7 @@ function rbac_version () {
 function version_gt() { 
     versions=$(for ver in "$@"; do ver=${ver#release-}; echo ${ver#v}; done)
     greaterVersion=${1#"release-"};  
+    greaterVersion=${1#"kubernetes-"};  
     greaterVersion=${greaterVersion#"v"}; 
     test "$(printf '%s' "$versions" | sort -V | head -n 1)" != "$greaterVersion"
 }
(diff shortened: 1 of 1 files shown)


deploy/util/deploy-hostpath.sh
```diff
@@ -218,6 +218,6 @@ while [ $(kubectl get pods 2>/dev/null | grep '^csi-hostpath.* Running ' | wc -l
 done
 
 # deploy snapshotclass
-echo "deploying snapshotclass"
-SNAPSHOTCLASS_PATH="${BASE_DIR}/snapshotter/csi-hostpath-snapshotclass.yaml"
-kubectl apply -f $SNAPSHOTCLASS_PATH
+echo "deploying snapshotclass based on snapshotter version"
+snapshotter_version="$(rbac_version "${BASE_DIR}/hostpath/csi-hostpath-snapshotter.yaml" csi-snapshotter false)"
+kubectl apply -f "https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/${snapshotter_version}/examples/kubernetes/snapshotclass.yaml"
```
msau42: this is sort of strange because the snapshotclass should live with the driver like storageclass does.

Another option I was thinking is to disable snapshots when the deployment version doesn't match the kubernetes version (before 1.17).  Because it was an alpha feature so there's no guarantees for cross version compatibility.

ggriffiths: Hmm ok, I'm unfamiliar with how to skip the snapshot tests. If we just don't apply a snapshotclass, will that cause the tests to be skipped?

msau42: while it was alpha, we only run the snapshot tests when the cluster and deployment version matched, and also only for the latest version.  So I think for logic < 1.17, we can make this assumption.

ggriffiths:
Closing and reopening to remove the pull-kubernetes-csi-csi-driver-host-path-1-14-on-kubernetes-1-14 expectation.

ggriffiths:
1.17 test flaked I believe.
/test pull-kubernetes-csi-csi-driver-host-path-1-17-on-kubernetes-1-17

PR #139, opened 2020-01-15
NodeVolumeUnpublish: tolerate repeated requests

### LINKED ISSUE #105: Idempotency: repeated Node Unpublish operations cause failure

okartau:
**What type of PR is this?**
/kind failing-test

NodeVolumeUnpublish handling has to be idempotent,
tolerating repeated requests.
That means, we can not attempt umount and return failure
without checking is that really mount point.

Fixes #105

Consider this as WIP: I intentionally replicated mount check code initially
in two case: blocks, and that's not meant to be final.
I made it like that to avoid semantic change to current switch block, which
already replicates code for two cases.
This switch does not have default (should it have?), i.e. we assume that
`VolAccessType` has to be one of two handled by case: blocks?
Does that mean we can bring mount check code to be before switch block?
But for that to be correct, we should probably consider adding default stmt in switch.
Note that two case: parts already have most of code identical
(we can use `targetPath` instead of `req.GetTargetPath()` in 2nd case)

```release-note
NONE
```

### DIFF
--- a/pkg/hostpath/nodeserver.go
+++ b/pkg/hostpath/nodeserver.go
@@ -204,25 +204,24 @@ func (ns *nodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
 		return nil, status.Error(codes.NotFound, err.Error())
 	}
 
-	switch vol.VolAccessType {
-	case blockAccess:
-		// Unmount and delete the block file.
+	// Unmount only if the target path is really a mount point.
+	if notMnt, err := mount.IsNotMountPoint(mount.New(""), targetPath); err != nil {
+		if !os.IsNotExist(err) {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+	} else if !notMnt {
+		// Unmounting the image or filesystem.
 		err = mount.New("").Unmount(targetPath)
 		if err != nil {
 			return nil, status.Error(codes.Internal, err.Error())
 		}
-		if err = os.RemoveAll(targetPath); err != nil {
-			return nil, status.Error(codes.Internal, err.Error())
-		}
-		glog.V(4).Infof("hostpath: volume %s has been unpublished.", targetPath)
-	case mountAccess:
-		// Unmounting the image
-		err = mount.New("").Unmount(req.GetTargetPath())
-		if err != nil {
-			return nil, status.Error(codes.Internal, err.Error())
-		}
-		glog.V(4).Infof("hostpath: volume %s/%s has been unmounted.", targetPath, volumeID)
 	}
+	// Delete the mount point.
+	// Does not return error for non-existent path, repeated calls OK for idempotency.
+	if err = os.RemoveAll(targetPath); err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+	glog.V(4).Infof("hostpath: volume %s has been unpublished.", targetPath)
 
 	if vol.Ephemeral {
 		glog.V(4).Infof("deleting volume %s", volumeID)


pohly:
> This switch does not have default (should it have?), i.e. we assume that
VolAccessType has to be one of two handled by case: blocks?

Better add a default.

> Note that two case: parts already have most of code identical
(we can use targetPath instead of req.GetTargetPath() in 2nd case)

Please try to avoid code duplication if possible.

okartau:
I added `default` case in that switch block, and that allowed to move common replicated parts out of switch-block. I also moved "check is mount point" test to happen before switch, without replication.

Although we discussed off-line with @pohly that more idempotency-safety via mutex locking could be added in same pass with this PR, I think it's simpler to take smaller steps, as this PR solely improves the ability to withstand repeated single-thread operations, as can be demonstrated by https://github.com/kubernetes-csi/csi-test/pull/229, once merged. 
The locking-based improvement is different story, it will improve test cases that dont exist yet, but can be added, making use of parallel requests. Thus, I propose to merge this PR first (enables to merge csi-test change adding repeated tests), and then move forward with locking changes.

pkg/hostpath/nodeserver.go
```diff
@@ -204,24 +204,27 @@ func (ns *nodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
 		return nil, status.Error(codes.NotFound, err.Error())
 	}
 
+	// Check if the target path is really a mount point. If its not a mount point do nothing
```
pohly: Nit: "If it's not a mount point do nothing."

pkg/hostpath/nodeserver.go
```diff
@@ -204,24 +204,27 @@ func (ns *nodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
...
+	if notMnt, err := mount.IsNotMountPoint(mount.New(""), targetPath); notMnt || err != nil && !os.IsNotExist(err) {
+		glog.V(4).Infof("hostpath: %s is not mount point, skip", targetPath)
+		return &csi.NodeUnpublishVolumeResponse{}, nil
```
pohly: If it is not mounted, we can only skip unmounting. We should not skip the rest of the function because a previous `NodeUnpublishVolume` might have been interrupted after unmounting and before removing the target path.

pkg/hostpath/nodeserver.go
```diff
@@ -204,24 +204,27 @@ func (ns *nodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
...
 		glog.V(4).Infof("hostpath: volume %s has been unpublished.", targetPath)
 	case mountAccess:
-		// Unmounting the image
-		err = mount.New("").Unmount(req.GetTargetPath())
-		if err != nil {
-			return nil, status.Error(codes.Internal, err.Error())
-		}
 		glog.V(4).Infof("hostpath: volume %s/%s has been unmounted.", targetPath, volumeID)
+	default:
+		return nil, status.Error(codes.Internal, fmt.Sprintf("unsupported access type %v", vol.VolAccessType))
```
pohly: In the unlikely case that we get here, we potentially did already some work (unmounting), but I think that's okay.

pkg/hostpath/nodeserver.go
okartau: in a closer look, we tell "if (not) mount point" twice which is one too many. Same message can be given without repeating same words, and in one statement. And combining that with another improvement idea from below, I change that comment to "Unmount only if the target path is really a mount point."

pkg/hostpath/nodeserver.go
okartau: For proper handling, we should check earlier that VolAccessType is one of supported types, and refuse to make any changes if not?

pkg/hostpath/nodeserver.go
okartau: if we only skip unmounting and proceed , we have to make sure repeated calling of os.RemoveAll is safe. It seems os.RemoveAll returns error if called with non-existent directory (is it so?), means we have to check for existence before calling?

pohly: Better check for the "not-exist" error after calling it and ignore that one.

okartau: yes, I also got idea to re-arrange those statements not to have everything packed on single line. (This line I copied from pmem-csi). But it improves readability to have it on different lines and check for err first, IMHO.

But also, I discovered that removal of directory in this functopn does not follow CSI spec which tells to remove the directory. Current code does it for VolAccessTyoe=Block only.

So what about:
- we remove the switch stmt so that there will be just os.RemoveAll.
- we add check for VolAccessType allowed values near start of function,
to avoid currently possible code path where some changes are made, then there is parameters check in the middle of function which can return without further changes.

okartau:
/retest
failure seems not related to change in this PR

pkg/hostpath/nodeserver.go
pohly: I don't think that's necessary. `vol.VolAccessType` really should be valid, as it was set earlier by the driver itself.

pkg/hostpath/nodeserver.go
pohly: > yes, I also got idea to re-arrange those statements not to have everything packed on single line. (This line I copied from pmem-csi). But it improves readability to have it on different lines and check for err first, IMHO.

I'm not sure I follow here.

`if err := <do something>; <check err here>` is the canonical way to handle errors. If the line gets too long, then simply spread it over multiple lines.

> Current code does it for VolAccessTyoe=Block only.

Yes, it should create (or try to create) the target directory in `NodePublishVolume` and remove it in `NodeUnpublishVolume`. It is a bug in Kubernetes that it creates the target path for filesystem volumes.

okartau: > I'm not sure I follow here.

What I meant is that instead of multiple ANDed and ORed login on same line:
```
if notMnt, err := mount.IsNotMountPoint(mount.New(""), targetPath); notMnt || err != nil && !os.IsNotExist(err) {
```

I prefer more straightforward structure:
notMnt, err := stmt
if err
.... and so on, having statement and checks separated for easier reading.
i.e. the code that is in PR now (I pushed new state close to my previous comment)

But what about my idea of removing switch stmt in middle. If you say we can trust volAccessType to be valid, does it mean then we can remove the switch altogether, without adding separate check closer to start of function ? And remove directory in all cases?

pohly: > I prefer more straightforward structure:
> notMnt, err := stmt
> if err

That's non-idiomatic Go. You can get the same readability with:
```
if notMnt, err := stmt;
    err ...
```

> If you say we can trust volAccessType to be valid, does it mean then we can remove the switch altogether, without adding separate check closer to start of function ?

Yes.

> And remove directory in all cases?

Yes.

okartau: I pushed new state addressing raised concerns. gofmt has own opinion how to format that "if" part so I used that. The switch stmt is gone, we always remove the directory. And we do not check for VolAccessType any more, we trust the type is valid.

pkg/hostpath/nodeserver.go
```diff
@@ -204,25 +204,23 @@ func (ns *nodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
...
+	if notMnt, err := mount.IsNotMountPoint(mount.New(""), targetPath); err != nil {
+		if !os.IsNotExist(err) {
 			return nil, status.Error(codes.Internal, err.Error())
 		}
-		glog.V(4).Infof("hostpath: volume %s has been unpublished.", targetPath)
-	case mountAccess:
+	} else if !notMnt {
 		// Unmounting the image
-		err = mount.New("").Unmount(req.GetTargetPath())
+		err = mount.New("").Unmount(targetPath)
```
pohly: The comment is no longer accurate. It should say "Unmounting the image or filesystem" now.

pkg/hostpath/nodeserver.go
```diff
@@ -204,25 +204,23 @@ func (ns *nodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
...
 		if err != nil {
 			return nil, status.Error(codes.Internal, err.Error())
 		}
-		glog.V(4).Infof("hostpath: volume %s/%s has been unmounted.", targetPath, volumeID)
 	}
+	// Delete the block file or mount point.
```
pohly: With "block file", do you mean the image file that the loop device is bound to? We don't delete that here.

I would just say "Delete the mount point." here, which is a correct statement, regardless whether that mount point is a directory (for filesystem mode) or file (for raw block).

okartau: I just re-used previous "block file" and added another part in that comment. Changed now as proposed, pushed.
One thing I notice when looking at latest code now, the recent change which took away "return OK" in "not mount point" case, likely shifted the potential idempotency failure point to next operation, os.RemoveAll which will fail if there is no mount point.
And I have not tested newest code against repeated test yet.
I need to do that before accept point of that code, I will do that locally soon, my guess is it now will fail at RemoveAll().
Partial reason why this is left unnoticed is that similar code on pmem-csi side does not check for return code of os.RemoveAll, i.e. tolerates failure silently, with end result being still OK.
The failure to remove non-existing directory is not really bad thing in that case.
But more clear is to check for return code and have code to handle this, commented.

okartau: or, what about another "code saving" approach: we do not check for return code and add comment why we do not?

okartau: no, I think correct way is to check. One of
- check first does path exist, if yes then RemoveAll, and also check for return code
- try RemoveAll and check for return code, if "no such path" then OK, otherwise error

pohly: The latter. Testing before some operation and then doing the operation is always a bit fishy. There have been security exploits because of this (https://hackernoon.com/time-of-check-to-time-of-use-toctou-a-race-condition-99c2311bd9fc).

This shouldn't be a problem here, but it's still better to follow best practices. It's also cheaper (= less syscalls), although again that doesn't matter here.

okartau: I tried current code and it works without error, i.e. os.RemoveAll does not return error on non-existing path
This is expected and documented on https://golang.org/pkg/os/ :
`If the path does not exist, RemoveAll returns nil (no error)`
Means, current code is good as is, needs comment stating above to make it even more clear (I will add a comment and push)

okartau:
/retest
failure seems not related to change in this PR

pohly:
/assign @msau42 

For approval.

msau42:
/approve
Thanks for fixing this!

PR #148, opened 2020-02-12
Add check for  volumecontent source in CreateVolume

Madhu-1:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

If the volume is already found during create volume request,we also need to check the volume content source id valid or not.
If the returning error message doesn't require formatting, use errors.New().

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/hostpath/controllerserver.go
+++ b/pkg/hostpath/controllerserver.go
@@ -127,23 +127,36 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 		// Since err is nil, it means the volume with the same name already exists
 		// need to check if the size of existing volume is the same as in new
 		// request
-		if exVol.VolSize >= int64(req.GetCapacityRange().GetRequiredBytes()) {
-			// existing volume is compatible with new request and should be reused.
-			// TODO (sbezverk) Do I need to make sure that volume still exists?
-			return &csi.CreateVolumeResponse{
-				Volume: &csi.Volume{
-					VolumeId:      exVol.VolID,
-					CapacityBytes: int64(exVol.VolSize),
-					VolumeContext: req.GetParameters(),
-					ContentSource: req.GetVolumeContentSource(),
-				},
-			}, nil
+		if exVol.VolSize < capacity {
+			return nil, status.Errorf(codes.AlreadyExists, "Volume with the same name: %s but with different size already exist", req.GetName())
 		}
-		return nil, status.Errorf(codes.AlreadyExists, "Volume with the same name: %s but with different size already exist", req.GetName())
+		if req.GetVolumeContentSource() != nil {
+			volumeSource := req.VolumeContentSource
+			switch volumeSource.Type.(type) {
+			case *csi.VolumeContentSource_Snapshot:
+				if volumeSource.GetSnapshot() != nil && exVol.ParentSnapID != volumeSource.GetSnapshot().GetSnapshotId() {
+					return nil, status.Error(codes.AlreadyExists, "existing volume source snapshot id not matching")
+				}
+			case *csi.VolumeContentSource_Volume:
+				if volumeSource.GetVolume() != nil && exVol.ParentVolID != volumeSource.GetVolume().GetVolumeId() {
+					return nil, status.Error(codes.AlreadyExists, "existing volume source volume id not matching")
+				}
+			default:
+				return nil, status.Errorf(codes.InvalidArgument, "%v not a proper volume source", volumeSource)
+			}
+		}
+		// TODO (sbezverk) Do I need to make sure that volume still exists?
+		return &csi.CreateVolumeResponse{
+			Volume: &csi.Volume{
+				VolumeId:      exVol.VolID,
+				CapacityBytes: int64(exVol.VolSize),
+				VolumeContext: req.GetParameters(),
+				ContentSource: req.GetVolumeContentSource(),
+			},
+		}, nil
 	}
 
 	volumeID := uuid.NewUUID().String()
-	path := getVolumePath(volumeID)
 
 	vol, err := createHostpathVolume(volumeID, req.GetName(), capacity, requestedAccessType, false /* ephemeral */)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/hostpath/controllerserver.go
```diff
@@ -152,19 +169,29 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 	glog.V(4).Infof("created volume %s at path %s", vol.VolID, vol.VolPath)
 
 	if req.GetVolumeContentSource() != nil {
-		contentSource := req.GetVolumeContentSource()
-		if snapshot := contentSource.GetSnapshot(); snapshot != nil {
-			err = loadFromSnapshot(snapshot.GetSnapshotId(), path)
-		}
-		if srcVolume := contentSource.GetVolume(); srcVolume != nil {
-			err = loadFromVolume(srcVolume.GetVolumeId(), path)
+
```
jsafrane: Why extra space?

pkg/hostpath/controllerserver.go
```diff
@@ -129,6 +129,24 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
...
+					if volumeSource.GetSnapshot() != nil && exVol.ParentSnapID != volumeSource.GetSnapshot().GetSnapshotId() {
+						err = fmt.Errorf("existing volume source snapshot id not matching")
+					}
+				case *csi.VolumeContentSource_Volume:
+					if volumeSource.GetVolume() != nil && exVol.ParentVolID != volumeSource.GetVolume().GetVolumeId() {
+						err = fmt.Errorf("existing volume source volume id not matching")
+					}
+				default:
+					err = fmt.Errorf("%v not a proper volume source", volumeSource)
+
```
jsafrane: nit: extra blank line

Madhu-1:
@jsafrane removed extra blank lines

pkg/hostpath/controllerserver.go
```diff
@@ -129,6 +129,23 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 		// request
 		if exVol.VolSize >= int64(req.GetCapacityRange().GetRequiredBytes()) {
 			// exisiting volume is compatible with new request and should be reused.
+
```
jsafrane: nit: extra blank line?

pkg/hostpath/controllerserver.go
```diff
@@ -129,6 +129,23 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
...
+						err = fmt.Errorf("existing volume source snapshot id not matching")
+					}
+				case *csi.VolumeContentSource_Volume:
+					if volumeSource.GetVolume() != nil && exVol.ParentVolID != volumeSource.GetVolume().GetVolumeId() {
+						err = fmt.Errorf("existing volume source volume id not matching")
+					}
+				default:
+					err = fmt.Errorf("%v not a proper volume source", volumeSource)
+				}
+				return nil, status.Error(codes.InvalidArgument, err.Error())
```
jsafrane: This always returns error even if  `VolumeContentSource` was fine and `err == nil`

pkg/hostpath/controllerserver.go
```diff
@@ -152,12 +168,20 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
...
 	if req.GetVolumeContentSource() != nil {
-		contentSource := req.GetVolumeContentSource()
-		if snapshot := contentSource.GetSnapshot(); snapshot != nil {
-			err = loadFromSnapshot(snapshot.GetSnapshotId(), path)
-		}
-		if srcVolume := contentSource.GetVolume(); srcVolume != nil {
-			err = loadFromVolume(srcVolume.GetVolumeId(), path)
+		volumeSource := req.VolumeContentSource
+		switch volumeSource.Type.(type) {
+		case *csi.VolumeContentSource_Snapshot, *csi.VolumeContentSource_Volume:
```
jsafrane: Should these be two separate cases? They share only one line.

pkg/hostpath/controllerserver.go
Madhu-1: yes correct, fixed now

pkg/hostpath/controllerserver.go
Madhu-1: split into a separate case

pkg/hostpath/controllerserver.go
```diff
@@ -129,6 +129,24 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
...
+					}
+				case *csi.VolumeContentSource_Volume:
+					if volumeSource.GetVolume() != nil && exVol.ParentVolID != volumeSource.GetVolume().GetVolumeId() {
+						err = fmt.Errorf("existing volume source volume id not matching")
+					}
+				default:
+					err = fmt.Errorf("%v not a proper volume source", volumeSource)
+				}
+				if err != nil {
+					return nil, status.Error(codes.InvalidArgument, err.Error())
```
jsafrane: In spirit of https://github.com/container-storage-interface/spec/pull/412, the first two cases should return AlreadyExists, while the `default` branch should return InvalidArgument. Moving `return` to the cases would be IMO better.

pkg/hostpath/controllerserver.go
```diff
@@ -129,6 +129,24 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 		// request
 		if exVol.VolSize >= int64(req.GetCapacityRange().GetRequiredBytes()) {
 			// exisiting volume is compatible with new request and should be reused.
+			if req.GetVolumeContentSource() != nil {
```
jsafrane: I am not happy with nested `if`s here.

IMO it would be better to do negative conditions:

```
if (size does not match){
  return AlreadyExists
}
if (source is not empty)
  ... return AlreadyExists / InvalidArgument on mismatch ...
}

// everything matches
return &csi.CreateVolumeResponse{}
```

pkg/hostpath/controllerserver.go
Madhu-1: updated code to return already exists

pkg/hostpath/controllerserver.go
Madhu-1: as I need to check 
* content source is empty or not
* if not need empty  check is it snapshot or volume or not 
* if it is a snapshot or volume check id is matching
nothing content source is neither snapshot nor volume return error

updated with `if` `else if` and `else` check

Madhu-1: now i got it, making changes

pkg/hostpath/controllerserver.go
```diff
@@ -127,23 +127,39 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
...
-			// TODO (sbezverk) Do I need to make sure that volume still exists?
-			return &csi.CreateVolumeResponse{
-				Volume: &csi.Volume{
-					VolumeId:      exVol.VolID,
-					CapacityBytes: int64(exVol.VolSize),
-					VolumeContext: req.GetParameters(),
-					ContentSource: req.GetVolumeContentSource(),
-				},
-			}, nil
+		if exVol.VolSize < int64(req.GetCapacityRange().GetRequiredBytes()) {
```
humblec: `capacity` is defined already for this , please use the same.

Madhu-1: done

Madhu-1:
@jsafrane @humblec PTAL

pkg/hostpath/controllerserver.go
```diff
@@ -127,23 +127,39 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
...
-			}, nil
+		if exVol.VolSize < capacity {
+			return nil, status.Errorf(codes.AlreadyExists, "Volume with the same name: %s but with different size already exist", req.GetName())
+		}
+		if req.GetVolumeContentSource() != nil {
+			volumeSource := req.VolumeContentSource
+			switch volumeSource.Type.(type) {
+			case *csi.VolumeContentSource_Snapshot:
+				if volumeSource.GetSnapshot() != nil && exVol.ParentSnapID != volumeSource.GetSnapshot().GetSnapshotId() {
+					err = status.Error(codes.AlreadyExists, "existing volume source snapshot id not matching")
```
jsafrane: nit: `return` right here, no need to pass `err` around. This applies to all cases + default.

Madhu-1: done.

jsafrane:
/lgtm
/approve
Thanks for the patience :-)

PR #149, opened 2020-02-12
Added corresponding ServiceAccount manifests

c3y1huang:
**What type of PR is this?**

/kind bug
/kind cleanup

Missing corresponding `ServiceAccount` manifests.

Fixes #

```release-note
NONE
```


Madhu-1:
these service accounts will be created as part of RBAC rules creation when you run deploy.sh from the deploy folder. we don't need to add them to host-path templates

PR #154, opened 2020-02-13
WIP: debug output for loop devices

pohly:
**What type of PR is this?**
/kind failing-test

Only for getting some more debug output about loop devices in Prow jobs. Don't merge.


pohly:
/close

The workaround seems to work (sic) now.

PR #157, opened 2020-02-17
Add support for block volume cloning

jsafrane:
Using `/bin/dd`.

/kind feature

```release-note
Added cloning support to raw block volumes.
```


jsafrane:
cc @j-griffith, this is needed for cloning tests
/assign @pohly @msau42

PR #161, opened 2020-02-28
Discover existing snapshots on disk

ashish-amarnath:
Signed-off-by: Ashish Amarnath <ashisham@vmware.com>

**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
This change lets this driver discover existing snapshots on disk and therefore improves testability of the scenario of "Importing an existing volume snapshot with Kubernetes"

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
Adds ability to discover and import existing on-disk tarballs as snapshots.
```

PR #162, opened 2020-03-02
Fixed block volume snapshots

jsafrane:
`tar xvfz <snapshot> -C <dst>` does not work for block volumes, because `<dst>` is not a directory but a block file.

Reworked to use `tar` only for filesystem snapshots and simple `cp`/`dd` for block snapshots. As consequence, snapshot suffix is now ".snap" instead of ".tgz".

/kind bug

Helps with #88726

```release-note
Fixed snapshots of block volumes.
```


jsafrane:
@msau42 @xing-yang @pohly, PTAL

msau42:
Do we support upgrading/restarting the hostpath driver and it being able to restore from previous snapshots? If so we should update the release note to say it's a breaking change

jsafrane:
Not yet, but see #161

PR #163, opened 2020-03-02
Remove snapshot e2e testing from alpha versions.

msau42:
**What type of PR is this?**
//kind failing-test

Disables snapshot e2es for 1.16 deployment.

prow.sh uses the e2e version that matches the k8s version. Skew testing
(driver version != k8s version) doesn't work for alpha features.

Fixes #

```release-note
NONE
```

### DIFF
--- a/deploy/kubernetes-1.15/test-driver.yaml
+++ b/deploy/kubernetes-1.15/test-driver.yaml
@@ -4,8 +4,6 @@
 
 StorageClass:
   FromName: true
-SnapshotClass:
-  FromName: true
 DriverInfo:
   Name: hostpath.csi.k8s.io
   Capabilities:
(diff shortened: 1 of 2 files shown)


msau42:
Failing testgrid: https://k8s-testgrid.appspot.com/sig-storage-csi-ci#1.16-on-1.17

xing-yang:
Should we also remove "snapshotter" folder and "hostpath/csi-hostpath-snapshotter.yaml" under kubernetes-1.15 and kubernetes-1.16?

msau42:
> Should we also remove "snapshotter" folder and "hostpath/csi-hostpath-snapshotter.yaml" under kubernetes-1.15 and kubernetes-1.16?

I think people may still want to deploy hostpath driver on older versions. We just can't test it in our e2es

PR #165, opened 2020-03-09
Update README.md to have latest snapshot API changes

### LINKED ISSUE #164: Update csi-driver-host-path README example to use latest images
README needs to be updated to use latest images.  In the current example, snapshot is still at v1alpha1.  Should be updated to use the beta version.

ggriffiths:
Signed-off-by: Grant Griffiths <grant@portworx.com>
**What type of PR is this?**
/kind documentation

README had the old snapshotter API as an example

Fixes #164

```release-note
NONE
```


README.md
```diff
@@ -183,31 +212,30 @@ An additional way to ensure the driver is working properly is by inspecting the
...
 Status:
   Attached:  true
 Events:      <none>
 ```
 
-
 ## Snapshot support
 
-Since volume snapshot is an alpha feature starting in Kubernetes v1.12, you need to enable feature gate called `VolumeSnapshotDataSource` in the Kubernetes.
+Since volume snapshot is an alpha feature from Kubernetes v1.12 through v1.16, you need to enable feature gate called `VolumeSnapshotDataSource` in Kubernetes for those versions.
```
xing-yang: Change to this:
Volume snapshot was introduced as an alpha feature in Kubernetes v1.12 and it was promoted to beta in Kubernetes v1.16.  Before v1.16, you need to manually enable feature gate called `VolumeSnapshotDataSource` in Kubernetes.

xing-yang:
> * use deploy/latest/deploy-hostpath.sh and add a note which explains which version was used for the output and examples in the rest of the readme.

I agree with this.

ggriffiths:
> The "Deployment" section starts with instructions to run `deploy/kubernetes-1.13/deploy-hostpath.sh`, which isn't in the repo anymore. Please either:
> 
> * replace with the version that you used or
> * use `deploy/latest/deploy-hostpath.sh` and add a note which explains which version was used for the output and examples in the rest of the readme.
> 
> We can't just use "latest" without additional information because if "latest" no longer works as described it would be hard to determine why.

Updated to include this.

README.md
ggriffiths: Fixed this.

README.md
```diff
@@ -13,54 +13,70 @@ the cluster as shown below for Kubernetes 1.13. This creates the deployment that
...
 rolebinding.rbac.authorization.k8s.io/csi-provisioner-role-cfg created
-kubectl apply -f https://raw.githubusercontent.com/kubernetes-csi/external-attacher/v1.0.1/deploy/kubernetes/rbac.yaml
+kubectl apply -f https://raw.githubusercontent.com/kubernetes-csi/external-attacher/v2.1.0/deploy/kubernetes/rbac.yaml
 serviceaccount/csi-attacher created
 clusterrole.rbac.authorization.k8s.io/external-attacher-runner created
 clusterrolebinding.rbac.authorization.k8s.io/csi-attacher-role created
 role.rbac.authorization.k8s.io/external-attacher-cfg created
 rolebinding.rbac.authorization.k8s.io/csi-attacher-role-cfg created
-kubectl apply -f https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/v1.0.1/deploy/kubernetes/rbac.yaml
+kubectl apply -f https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/v2.0.0/deploy/kubernetes/csi-snapshotter/rbac-csi-snapshotter.yaml
```
xing-yang: We should actually use v2.0.1 because it contains a fix for go module.

README.md
```diff
@@ -13,54 +13,70 @@ the cluster as shown below for Kubernetes 1.13. This creates the deployment that
...
 service/csi-hostpath-provisioner created
 statefulset.apps/csi-hostpath-provisioner created
-   deploy/kubernetes-1.13/hostpath/csi-hostpath-snapshotter.yaml
-        using           image: quay.io/k8scsi/csi-snapshotter:v1.0.1
+   deploy/kubernetes-1.17/hostpath/csi-hostpath-resizer.yaml
+        using           image: quay.io/k8scsi/csi-resizer:v0.4.0
+service/csi-hostpath-resizer created
+statefulset.apps/csi-hostpath-resizer created
+   deploy/kubernetes-1.17/hostpath/csi-hostpath-snapshotter.yaml
+        using           image: quay.io/k8scsi/csi-snapshotter:v2.0.0
```
xing-yang: We should use v2.0.1

xing-yang:
Can you also update the snapshot images to v2.0.1 in the script?

README.md
ggriffiths: Fixed - running the same commands from 2.0.1 now to double check the same functioning.

README.md
ggriffiths: Fixed this

README.md
```diff
@@ -6,61 +6,94 @@ This repository hosts the CSI Hostpath driver and all of its build and dependent
 - Kubernetes cluster
 - Running version 1.13 or later
 - Access to terminal with `kubectl` installed
+- Snapshot controller and snapshot CRDs applied (see below)
```
xing-yang: This only works for k8s version 1.17 and higher.

@msau42 should we have separate sections for "k8s 1.17 and higher" and "version lower than 1.17" (as we still have 1.16 and 1.15 deployment scripts).

README.md
```diff
@@ -6,61 +6,94 @@ This repository hosts the CSI Hostpath driver and all of its build and dependent
...
+
+## Snapshot controller and CRDs setup
```
xing-yang: This section is only for k8s 1.17 and higher.

README.md
msau42: Yes I think that makes sense.

Also, in general, I think this readme is quite long and difficult to find specific topics. Can we potentially break it up a page per example, under a docs/ folder? And make the top-level readme more of TOC?

ggriffiths: Makes sense - I'll update the README to point to point to these two new pages. Will save the old example output for the <1.17 version

ggriffiths: Ok, I've broken up the deployment docs under 
docs/deploy-latest.md
docs/deploy-pre-1.17.md

For the pre-1.17 docs, I copied the old deployment instructions but  reran with deploy/kubernetes-1.16/deploy-hostpath.sh and updated that page.

ggriffiths: oh, I just re-read your comment. I'll break the examples down into separate pages too.

ggriffiths: Ok, broke README.md down to the following files:
```
docs/deploy-latest.md
docs/deploy-pre-1.17.md
docs/example-snasphots.md
docs/example-ephemeral.md
```

docs/example-ephemeral.md
```diff
@@ -0,0 +1,55 @@
+## Inline ephemeral support
+The CSI Hostpath driver (starting with version 1.2.0) now includes support for inline ephemeral volumes. This means that a volume can be specified directly inside a pod spec without the need to use a persistent volume object.
```
xing-yang: Can you add a line here:
Inline ephemeral support was introduced as an alpha feature in Kubernetes 1.15 and was promoted to beta in 1.16.

docs/example-snapshots.md
```diff
@@ -0,0 +1,156 @@
...
+
+> For Kubernetes < v1.17 
+>
+>  - `$ kubectl apply -f examples/csi-snapshot.yaml`
+> ```
+> $ kubectl apply -f examples/csi-snapshot.yaml
+> volumesnapshot.snapshot.storage.k8s.io/new-snapshot-demo created
+> ```
+>
+> For Kubernetes >= v1.17
```
xing-yang: Can you also split example-snapshots.md into two, one for examples < 1.17 and one for 1.17+

README.md
```diff
@@ -6,403 +6,17 @@ This repository hosts the CSI Hostpath driver and all of its build and dependent
...
-      csi:
-        driver: hostpath.csi.k8s.io
-``` 
-
-> See sample YAML file [here](./examples/csi-app-inline.yaml).
-
-Notice the CSI driver is now specified directly in the container spec inside the `volumes:` block.  You can use the [same steps as above][Confirm Hostpath driver works] 
-to verify that the volume has been created and deleted (when the pod is removed).
+Deployment varies depending on the Kubernetes version your cluster is running:
+- [Deployment for Kubernetes 1.17+](docs/deploy-latest.md)
```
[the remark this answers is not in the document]
ggriffiths: Fixed

docs/deploy-latest.md
```diff
@@ -0,0 +1,248 @@
...
+For Kubernetes 1.17+, some initial cluster setup is required to install the following:
+- CSI VolumeSnapshot beta CRDs
+- Snapshot Controller
+
+Run the following commands to install these components: 
+```shell
+# Change to the latest supported snapshotter version
+$ SNAPSHOTTER_VERSION=v2.0.1
+
+# Apply snapshotter CRDs
```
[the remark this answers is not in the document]
ggriffiths: Fixed

docs/deploy-latest.md
```diff
@@ -0,0 +1,248 @@
...
+$ kubectl apply -f https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/${SNAPSHOTTER_VERSION}/config/crd/snapshot.storage.k8s.io_volumesnapshots.yaml
+
+# Create snapshot controller
+$ kubectl apply -f https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/${SNAPSHOTTER_VERSION}/deploy/kubernetes/snapshot-controller/rbac-snapshot-controller.yaml
+$ kubectl apply -f https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/${SNAPSHOTTER_VERSION}/deploy/kubernetes/snapshot-controller/setup-snapshot-controller.yaml
+```
+
+## Deployment
+The easiest way to test the Hostpath driver is to run the `deploy-hostpath.sh` script for the Kubernetes version used by
+the cluster as shown below for Kubernetes 1.13. This creates the deployment that is maintained specifically for that
```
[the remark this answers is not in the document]
ggriffiths: Fixed - the output was from 1.17

ggriffiths:
Addressed feedback @xing-yang

README.md
```diff
@@ -6,403 +6,19 @@ This repository hosts the CSI Hostpath driver and all of its build and dependent
...
-      csi:
-        driver: hostpath.csi.k8s.io
-``` 
-
-> See sample YAML file [here](./examples/csi-app-inline.yaml).
-
-Notice the CSI driver is now specified directly in the container spec inside the `volumes:` block.  You can use the [same steps as above][Confirm Hostpath driver works] 
-to verify that the volume has been created and deleted (when the pod is removed).
+Deployment varies depending on the Kubernetes version your cluster is running:
+- [Deployment for Kubernetes 1.17+](docs/deploy-1.17-and-later.md)
```
xing-yang: nit: s/1.17+/1.17 and later

README.md
```diff
@@ -6,403 +6,19 @@ This repository hosts the CSI Hostpath driver and all of its build and dependent
...
+- [Deployment for Kubernetes 1.16 or earlier](docs/deploy-pre-1.17.md)
```
xing-yang: s/1.16 or earlier/1.16 and earlier

docs/deploy-1.17-and-later.md
```diff
@@ -0,0 +1,248 @@
...
+
+# Create snapshot controller
+$ kubectl apply -f https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/${SNAPSHOTTER_VERSION}/deploy/kubernetes/snapshot-controller/rbac-snapshot-controller.yaml
+$ kubectl apply -f https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/${SNAPSHOTTER_VERSION}/deploy/kubernetes/snapshot-controller/setup-snapshot-controller.yaml
+```
+
+## Deployment
+The easiest way to test the Hostpath driver is to run the `deploy-hostpath.sh` script for the Kubernetes version used by
+the cluster as shown below for Kubernetes 1.17. This creates the deployment that is maintained specifically for that
+release of Kubernetes. However, other deployments may also work. For details see the individual READMEs.
```
xing-yang: This sentence looks odd: "However, other deployments may also work. For details see the individual READMEs."
There is only one README for 1.17 and later, so I don't know what are the other individual READMEs.
Probably better just to drop this sentence.

docs/deploy-pre-1.17.md
```diff
@@ -0,0 +1,210 @@
+## Deployment
+The easiest way to test the Hostpath driver is to run the `deploy-hostpath.sh` script for the Kubernetes version used by
+the cluster as shown below for Kubernetes 1.16. This creates the deployment that is maintained specifically for that
+release of Kubernetes. However, other deployments may also work. For details see the individual READMEs.
```
xing-yang: Same comment here: drop "For details see the individual READMEs." as there is only 1 readme for pre-1.17.

docs/deploy-1.17-and-later.md
ggriffiths: Removed this

docs/deploy-pre-1.17.md
ggriffiths: Removed this

README.md
ggriffiths: Fixed

README.md
ggriffiths: Fixed

README.md
```diff
@@ -6,403 +6,19 @@ This repository hosts the CSI Hostpath driver and all of its build and dependent
 - Kubernetes cluster
 - Running version 1.13 or later
 - Access to terminal with `kubectl` installed
+- For Kubernetes 1.17 or later, the VolumeSnapshot beta CRDs and Snapshot Controller must be installed (see Kubernetes 1.17+ deployment instructions)
```
msau42: clarify this should be installed as part of Kubernetes cluster deployment.

docs/deploy-1.17-and-later.md
```diff
@@ -0,0 +1,248 @@
+## Cluster setup
+For Kubernetes 1.17+, some initial cluster setup is required to install the following:
```
msau42: Clarify that these steps should only be run if your cluster hasn't already packaged these components in. And maybe add some commands to easily check if these resources already exist.

README.md
ggriffiths: Fixed

docs/deploy-1.17-and-later.md
ggriffiths: Updated with some commands to check for the CRDs and Snapshot controller

docs/deploy-1.17-and-later.md
```diff
@@ -0,0 +1,269 @@
...
+$ kubectl get volumesnapshotcontents.snapshot.storage.k8s.io
+```
+If any of these commands return the following error message, you must install the corresponding CRD:
+```
+error: the server doesn't have a resource type "volumesnapshotclasses"
+```
+
+Next, check if any pods are running the snapshot-controller image:
+```
+$ kubectl get pods --all-namespaces -o=jsonpath='{range .items[*]}{"\n"}{range .spec.containers[*]}{.image}{", "}{end}{end}' | grep snapshot-controller
```
msau42: hm this doesn't work on managed distros where you can't see the master pods. Is there another way to check? maybe add a note that if you're running on a managed k8s service, this step may not work, and maybe you'll have to just use the presence of the crds as an indication.

ggriffiths: Ah yeah, good point. I'm not sure of any other way. @xing-yang any ideas?

I'll add a note to clarify this.

msau42:
/lgtm
/approve
Thanks!

PR #184, opened 2020-06-15
Do not use latest version of image against older version of rbac rules

gnufied:
Using canary image with older version of rbac rules do not work for resizer.

The job in question - https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-csi/csi-driver-host-path/csi-driver-host-path-config.yaml#L814 

I do not see how this can be made to work either without black listing csi-resizer here or by setting `UPDATE_RBAC_RULES` to `true` in linked job.

/assign @pohly @msau42

### DIFF
--- a/deploy/kubernetes-1.16/canary-blacklist.txt
+++ b/deploy/kubernetes-1.16/canary-blacklist.txt
@@ -3,3 +3,4 @@
 csi-snapshotter
 csi-attacher
 csi-provisioner
+csi-resizer


msau42:
/lgtm
/approve

Make sure in the next release notes for resizer to mention the new rbacs.

pohly:
> I do not see how this can be made to work either without black listing csi-resizer here or by setting UPDATE_RBAC_RULES to true in linked job.

The other solution is to accept that the upcoming csi-resizer release will require a major version bump because the newer image no longer works in existing, unmodified deployments. The "canary" jobs are testing for that, so that ci-kubernetes-csi-canary-on-kubernetes-1-16 failing is in that sense intentional.

After a major version bump, we typically stop testing on older Kubernetes releases, so the job could be disabled.

pohly:
So yes, this PR looks correct to me.

gnufied:
@pohly @msau42 ack to both of your comments. I have it noted for next resizer release - https://github.com/orgs/kubernetes-csi/projects/12#card-40136501

PR #196, opened 2020-08-13
Remove services from Sidecar yaml

### LINKED ISSUE #170: Remove service from hostpath specs
They're unnecessary and actually a random port is specified, which may not be desirable.

We got confirmation from sig-apps in https://github.com/kubernetes/kubernetes/issues/69608 that this will be supported. StatefulSet is v1 now, so they cannot change the behavior without it being breaking change.

/help
/kind cleanup

Kartik494:
**What type of PR is this?**

> /kind cleanup

Removing Service Spec from sidecar yaml (csi-hostpath-attacher.yaml,csi-hostpath-provisioner.yaml,csi-hostpath-snapshotter.yaml,csi-hostpath-resizer.yaml ) for Kubernetes-1.16,Kubernetes-1.17,Kubernetes-1.18

Fixes #170 

```
NONE
```

### DIFF
--- a/deploy/kubernetes-1.16/hostpath/csi-hostpath-attacher.yaml
+++ b/deploy/kubernetes-1.16/hostpath/csi-hostpath-attacher.yaml
@@ -1,17 +1,3 @@
-kind: Service
-apiVersion: v1
-metadata:
-  name: csi-hostpath-attacher
-  labels:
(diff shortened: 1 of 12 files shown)


Kartik494:
assign @jsafrane @vladimirvivien Could you please take a look on this

PR #210, opened 2020-10-18
feature: support external-health-monitor

fengzixu:
**What type of PR is this?**

/kind feature
Implementing the `ListVolumes`, `GetVolume` and `NodeStats` interfaces to support external-health-monitor project

Fixes https://github.com/kubernetes-csi/external-health-monitor/issues/21

```release-note
support external-health-monitor feature in hostpath driver
```


xing-yang:
This is a new feature.  Please add release notes.

fengzixu:
The implementation of support external-health-monitor in host-path-driver has been finished. I have tested this PR in my personal env. But I'm not sure If I need to change the orchestration file in deploy directory to add related argument of deploying external-health-monitor-controller and agent.

@xing-yang After you reviewed the implementation, please tell me if i need to do it

xing-yang:
Add a release note.

xing-yang:
Can you take a look at the CI failures?  Did you do `go mod vendor` and `go mod tidy`?

pkg/hostpath/hostpath.go
```diff
@@ -173,6 +195,32 @@ func (hp *hostPath) Run() {
...
+		hostPathVolumes[pv.Spec.CSI.VolumeHandle] = hostPathVolume{
+			VolName:       pv.Name,
+			VolID:         pv.Spec.CSI.VolumeHandle,
+			VolSize:       pv.Spec.Capacity.Storage().Value(),
+			VolPath:       getVolumePath(pv.Spec.CSI.VolumeHandle),
+			VolAccessType: volAccessType,
+		}
+	}
+
+	glog.V(3).Infof("The existing volume list is: %+v\n", hostPathVolumes)
```
xing-yang, line 220: Remove "\n".

deploy/kubernetes-1.17/hostpath/csi-hostpath-driverinfo.yaml
```diff
@@ -4,9 +4,6 @@ metadata:
   name: hostpath.csi.k8s.io
 spec:
   # Supports persistent and ephemeral inline volumes.
-  volumeLifecycleModes:
-  - Persistent
-  - Ephemeral
   # To determine at runtime which mode a volume uses, pod info and its
```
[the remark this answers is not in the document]
fengzixu: oh. Sorry. My bad. Let me revert it

pkg/hostpath/healthcheck.go
```diff
@@ -0,0 +1,201 @@
...
+	spExist, err := checkSourcePathExist(volumeHandle)
+	if err != nil {
+		return false, err.Error()
+	}
+
+	if !spExist {
+		return false, "The source path of the volume doesn't exist"
+	}
+
+	mpExist, err := checkMountPointExist(volumeHandle)
```
xing-yang, line 173: I think only this mountpoint check is relevant for NodeGetVolumeStats.  
Other checks are for ListVolumes() and GetVolume() from the controller side.

pkg/hostpath/nodeserver.go
```diff
@@ -294,12 +296,45 @@ func (ns *nodeServer) NodeGetCapabilities(ctx context.Context, req *csi.NodeGetC
...
 }
 
 func (ns *nodeServer) NodeGetVolumeStats(ctx context.Context, in *csi.NodeGetVolumeStatsRequest) (*csi.NodeGetVolumeStatsResponse, error) {
-	return nil, status.Error(codes.Unimplemented, "")
+	volume, ok := hostPathVolumes[in.GetVolumeId()]
+	if !ok {
+		return nil, status.Error(codes.NotFound, "The volume not found")
+	}
+
+	healthy, msg := doHealthCheck(in.GetVolumeId())
```
xing-yang, line 316: Looks like you are calling the same `doHealthCheck` for NodeGetVolumeStats, ListVolumes, and GetVolume.  That will produce duplicate events on PVCs and Pods.  Health check from controller side should be separate from node side so there should not be duplicate events.
For NodeGetVolumeStats, we should only need to check things like the mount condition.

xing-yang:
Did you run "go mod vendor" and "go mode tidy" after removing calls to k8s api server?

fengzixu:
> Did you run "go mod vendor" and "go mode tidy" after removing calls to k8s api server?

Sure

pkg/hostpath/nodeserver.go
fengzixu: Honestly, it’s hard to separate health check cases in the host-path driver. Because host-path-driver runs all of CSI components on the same node.

But, to avoid duplicate the event, it's one of the reasons we need to separate them.

deploy/kubernetes-1.17/hostpath/csi-hostpath-plugin.yaml
```diff
@@ -34,7 +34,37 @@ spec:
...
+            - "--v=5"
+            - "--csi-address=$(ADDRESS)"
+            - "--leader-election"
+          env:
+            - name: ADDRESS
+              value: /csi/csi.sock
+          imagePullPolicy: "IfNotPresent"
+          volumeMounts:
+            - name: socket-dir
+              mountPath: /csi
```
xing-yang, line 67: CI will not pass with these changes in hostpath deployment.

An older version of csi-test is deployed which does not support Volume_Condition added to the newer csi-test.
https://github.com/kubernetes-csi/csi-release-tools/blob/master/prow.sh#L230
https://github.com/kubernetes-csi/csi-test/commit/5421d9f3c37be3b95b241b44a094a3db11bee789

We'll need to make changes in CI before making changes in hostpath deployment file.

pkg/hostpath/controllerserver.go
```diff
@@ -352,7 +423,7 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 	snapshot.Id = snapshotID
 	snapshot.VolID = volumeID
 	snapshot.Path = file
-	snapshot.CreationTime = *creationTime
+	snapshot.CreationTime = creationTime
```
xing-yang, line 426: Do you have to change CreateSnapshot?  This is not related to volume health.

pkg/hostpath/healthcheck.go
```diff
@@ -0,0 +1,201 @@
...
+	"strings"
+
+	"github.com/golang/glog"
+	fs "k8s.io/kubernetes/pkg/volume/util/fs"
+)
+
+const (
+	// sourcePathPrefix should be specified by the csi-data-dir volume.
+	// You can change it by modifying the volume path of csi-data-dir volume.
+	sourcePathPrefix          = "/csi-data-dir/"
```
xing-yang, line 18: Isn't this already defined in hostpath.go?  Can you just use that?

pkg/hostpath/controllerserver.go
fengzixu: Because `go mod tidy` upgrade the grpc dependency. In the new version of grpc, it doesn't permit we assign a copy value.

pkg/hostpath/healthcheck.go
fengzixu: Sure

deploy/kubernetes-1.17/hostpath/csi-hostpath-plugin.yaml
fengzixu: https://github.com/kubernetes-csi/csi-release-tools/pull/117 I gave a PR to the release-tools repo.After it was merged, I think I need to use git subtree command to update the release-tools in the host-path-driver. Right?

fengzixu:
I change the prow.sh to upgrade the csi-test to v4.0.2. It seems like all of e2e tests have been passed. But I don't know why the robot still reports it as failed @xing-yang

fengzixu:
Oh. It seems like somes tests are timeout

release-tools/prow.sh
```diff
@@ -227,8 +227,9 @@ configvar CSI_PROW_E2E_IMPORT_PATH "k8s.io/kubernetes" "E2E package"
 # of the cluster. The alternative would have been to (cross-)compile csi-sanity
 # and install it inside the cluster, which is not necessarily easier.
 configvar CSI_PROW_SANITY_REPO https://github.com/kubernetes-csi/csi-test "csi-test repo"
-configvar CSI_PROW_SANITY_VERSION 5421d9f3c37be3b95b241b44a094a3db11bee789 "csi-test version" # latest master
-configvar CSI_PROW_SANITY_IMPORT_PATH github.com/kubernetes-csi/csi-test "csi-test package"
+configvar CSI_PROW_SANITY_VERSION v4.0.2 "csi-test version" # latest master
+t
```
pohly, line 231: How did this pass testing?

NickrenREN:
hostpath unit task still fails:
```
make: *** [release-tools/build.make:259: test-shellcheck] Error 1
make: Target 'test' not remade because of errors.
ERROR: 'make test' failed
WARNING: 'make test' failed, proceeding anyway
```

fengzixu:
> hostpath unit task still fails:
> 
> ```
> make: *** [release-tools/build.make:259: test-shellcheck] Error 1
> make: Target 'test' not remade because of errors.
> ERROR: 'make test' failed
> WARNING: 'make test' failed, proceeding anyway
> ```

@NickrenREN I already fixed it in this PR https://github.com/kubernetes-csi/csi-release-tools/pull/128.
Actually, we cannot directly modify prow.sh in this PR, it is just for testing if other e2e test can be passed.

We need to merge https://github.com/kubernetes-csi/csi-release-tools/pull/128 first. And submit a new PR to host-path repo to update prow.sh by git subtree command. Finally, we can rebase this PR to resolve all of test failure

NickrenREN:
>@NickrenREN I already fixed it in this PR kubernetes-csi/csi-release-tools#128.
Actually, we cannot directly modify prow.sh in this PR, it is just for testing if other e2e test can be passed.
We need to merge kubernetes-csi/csi-release-tools#128 first. And submit a new PR to host-path repo to update prow.sh by git subtree command. Finally, we can rebase this PR to resolve all of test failure

get it, thanks

NickrenREN:
thanks for this. 

/lgtm

xing-yang:
Under the section "Special notes for your reviewer" in the PR description, can you add more details on what are checked and what are reported as abnormal volume condition in the health monitor controller and agent respectively?

pohly:
Updating E2E testing in Kubernetes from csi-driver-host-path 1.4.0 to 1.6.2 causes jobs to fail and the code from this PR seems to be involved, see https://github.com/kubernetes/kubernetes/pull/100637#discussion_r605553855.

pohly:
Do we have any test for the `discoveryExistingVolumes` function?

Dockerfile
```diff
@@ -4,6 +4,6 @@ LABEL description="HostPath Driver"
 ARG binary=./bin/hostpathplugin
 
 # Add util-linux to get a new version of losetup.
-RUN apk add util-linux
+RUN apk add util-linux && apk update && apk upgrade
```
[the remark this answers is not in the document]
fengzixu: When I implement this feature, I found I need run findmnt command with `json`argument. It was supported in newer version. So, I update it

cmd/hostpathplugin/main.go
```diff
@@ -63,5 +63,10 @@ func handle() {
 		fmt.Printf("Failed to initialize driver: %s", err.Error())
 		os.Exit(1)
 	}
-	driver.Run()
+
+	if err := driver.Run(); err != nil {
+		fmt.Printf("Failed to run driver: %s", err.Error())
+		os.Exit(1)
+
+	}
```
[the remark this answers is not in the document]
fengzixu: An error may be returned from that function, just checking it right?

pkg/hostpath/hostpath.go
```diff
@@ -161,8 +165,56 @@ func discoverExistingSnapshots() {
...
+			break
+		}
+	}
+
+	// getting existing volumes based on the mount point infos.
+	// It's a temporary solution to recall volumes.
+	for _, pv := range mountInfosOfPod.ContainerFileSystem {
+		if !strings.Contains(pv.Target, csiSignOfVolumeTargetPath) {
+			continue
+		}
```
pohly, line 198: How does this code distinguish between mounts from some other CSI driver and the mounts of this hostpath driver instance?

pohly: I think it doesn't.

Both `podVolumeTargetPath = /var/lib/kubelet/pods` and `csiSignOfVolumeTargetPath = kubernetes.io~csi/pvc` are shared with all other CSI driver instances.

pohly: Besides the obvious failure of adding volumes that aren't actually managed by this driver, we also get a race condition that leads to the "failed to get capacity info: no such file or directory" error from https://github.com/kubernetes/kubernetes/pull/100637#discussion_r605553855

- one CSI driver starts listing mounts
- another CSI driver concurrently unmounts
- the first CSI driver tries to get capacity information which then fails

Does that make sense?

pohly: How is `discoveryExistingVolumes` related to volume health checks?

I understand the desire to support container restarts. I just don't understand how it is related to health checks.

Regarding container restarts: the information restored is incomplete (for example, the more recently added nominal capacity of a volume cannot be discovered). Instead of making this code more complex, can we simply rip it out and go back to the state from the hostpath 1.4 release, i.e. the driver must not be restarted?

pohly: Same with discovering snapshots?

xing-yang: We need to discover existing volumes to implement ListVolumes().

I think we need to check driver name to make sure this is a PV provisioned by CSI hostpath driver.

If we get "no such file or directory" error, that means the volume does not exist any more. I think we should not return error in this case.

pohly: > We need to discover existing volumes to implement ListVolumes().

Why can't you use the in-memory list of known volumes for that?

> I think we need to check driver name to make sure this is a PV provisioned by CSI hostpath driver.

Probably something like that.

> If we get "no such file or directory" error, that means the volume does not exist any more. I think we should not return error in this case.

Once the driver has correctly identified a mount point as its own, the volume should not suddenly disappear. Returning an error when that happens seems appropriate to me.

pohly: > > We need to discover existing volumes to implement ListVolumes().
> 
> Why can't you use the in-memory list of known volumes for that?

In fact, that's exactly how `ListVolumes` was implemented...

xing-yang: discoveryExistingVolumes() is the function that builds the in-memory list.  It is called when host path driver is initialized.

pohly: Yes, I know. But why do we need to support driver restarts?

That is orthogonal to supporting health checks.

xing-yang: When testing upgrade for moving snapshot from v1beta1 to v1, for example, driver restart will happen.  So we need to handle driver restarts.

I'm open to different ways of implementing this.

pohly: So you needed that for some other purpose. Now it makes more sense to me.

I started implementing this, but won't have time to finish it. If someone else wants to take a stab at it, feel free.

IMHO the code which Volume and Snaphshot structs and the corresponding maps should be in its own package, with get/add/update/delete functions that have to be used to make changes. Currently it's in the same package and I already found code which directly manipulated the maps instead of using the existing helper functions.

Then in that package, we can dump the entire struct into one file to store volumes and snapshots. That removes a whole lot of custom code, including the one which we discuss here. Combine that with a configurable data dir and we can add unit tests for this...

xing-yang: @fengzixu can you take a stab at this?

fengzixu: Sure. Let me try

Issue #212, opened 2020-10-21
destroying example deployment with a script


irbekrm:
Hi!

Thanks for this great project!

I was following the example linked from Readme and using the script in `deploy/util/deploy-hostpath.sh` to deploy the driver and the sidecars and it was really convenient, but if I want to clean up afterwards it takes a bit of time, especially to find the RBAC resources the script deployed- I was wondering if it wouldn't be nice to also be able to afterwards delete all the deployed resources with a corresponding `destroy-hostpath.sh` script?

I would be interested to write the script if you think it could be useful.

pohly:
I think that would be useful.

Labeling the resources and then deleting those by selecting them via that label would be a good way to keep the script generic.

aayushrangwala:
@pohly  I have came across similar requirement. I believe we should make the deploy script to be generic and pass the operator as an arg. Like `hostpath-driver.sh install ` and `hostpath-driver.sh delete` because we are evaluating paths and versions in the script so even at the time of destroying we won't need to re-evaluate them since rbacs doesn't have labels at the moment, we cannot just directly delete using labels.

Or the other option I think is we should create another script, delete all those things like sts which have labels like `csi-provisioner`, `csi-attacher` etc and then just evaluate the rbac versions and delete them using files directly to maintain consistency with what we apply to what we delete

pohly:
> since rbacs doesn't have labels at the moment, we cannot just directly delete using labels.

RBAC objects have an `ObjectMeta` and thus can have labels. That they currently don't have them is something that we can and should change once we add deletion by label.

> delete them using files directly to maintain consistency with what we apply to what we delete

It may be useful to delete objects created by a different version of the scripts, for example when it isn't even known anymore which exact version was used for deploying, or simply to keep it easy to use.

When deleting exactly what would be installed, we may miss deleting of things that no longer get installed at all or were renamed.

aayushrangwala:
right!, make sense. thanks. I will create seperate PRs for adding labels to all the rbac-files and also the resources which dont have right now (probably for the older versions also) and then will just add a PR for deleting script in this repo.

pohly:
/remove-lifecycle stale

@aayushrangwala do you still intend to work on this?

PR #216, opened 2020-11-09
added a destroy-script for the resources installed

aayushrangwala:
- added labels to the resources

**What type of PR is this?**
/kind cleanup

Need a cleanup script to delete all the installed resources

Fixes https://github.com/kubernetes-csi/csi-driver-host-path/issues/212

NONE

```release-note
All the csi resources from k8s 1.17 will be having a common label `partof: csi-hostpath-driver` which will be used to collectively identify all the resources under csi installation and also can be used to delete them in bulk
```

### DIFF
--- a/deploy/kubernetes-1.20/hostpath/csi-hostpath-plugin.yaml
+++ b/deploy/kubernetes-1.20/hostpath/csi-hostpath-plugin.yaml
@@ -7,7 +7,10 @@ apiVersion: v1
 metadata:
   name: csi-hostpathplugin
   labels:
-    app: csi-hostpathplugin
+    app.kubernetes.io/instance: hostpath.csi.k8s.io
+    app.kubernetes.io/part-of: csi-driver-host-path
+    app.kubernetes.io/name: csi-hostpathplugin
+    app.kubernetes.io/component: plugin
 spec:
   selector:
     app: csi-hostpathplugin
@@ -19,6 +22,11 @@ kind: StatefulSet
 apiVersion: apps/v1
 metadata:
   name: csi-hostpathplugin
+  labels:
+    app.kubernetes.io/instance: hostpath.csi.k8s.io
+    app.kubernetes.io/part-of: csi-driver-host-path
+    app.kubernetes.io/name: csi-hostpathplugin
+    app.kubernetes.io/component: plugin
 spec:
   serviceName: "csi-hostpathplugin"
   # One replica only:
@@ -28,11 +36,17 @@ spec:
   replicas: 1
   selector:
     matchLabels:
-      app: csi-hostpathplugin
+      app.kubernetes.io/instance: hostpath.csi.k8s.io
+      app.kubernetes.io/part-of: csi-driver-host-path
+      app.kubernetes.io/name: csi-hostpathplugin
+      app.kubernetes.io/component: plugin
   template:
     metadata:
       labels:
-        app: csi-hostpathplugin
+        app.kubernetes.io/instance: hostpath.csi.k8s.io
+        app.kubernetes.io/part-of: csi-driver-host-path
+        app.kubernetes.io/name: csi-hostpathplugin
+        app.kubernetes.io/component: plugin
     spec:
       serviceAccountName: csi-external-health-monitor-controller
       containers:
--- a/deploy/kubernetes-distributed/destroy.sh
+++ b/deploy/kubernetes-distributed/destroy.sh
@@ -1,0 +1,34 @@
+#!/usr/bin/env bash
+# Copyright 2021 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This script performs the step to destroy the hostpath plugin driver
+# completely by checking all the resources.
+# This should be in sync with the deploy script based on any future additions/deletions
+# of the resources
+
+# The script assumes that kubectl is available on the OS path
+# where it is executed.
+
+set -e
+set -o pipefail
+
+# Deleting all the resources installed by the deploy-script as the part of csi-hostpath-driver.
+# Every resource in the driver installation has the label representing the installation instance.
+# Using app.kubernetes.io/instance: hostpath.csi.k8s.io and app.kubernetes.io/part-of: csi-driver-host-path labels to identify the installation set
+#
+# kubectl maintains a few standard resources under "all" category which can be deleted by using just "kubectl delete all"
+# and other resources such as roles, clusterrole, serivceaccount etc needs to be deleted explicitly
+kubectl delete all --all-namespaces -l app.kubernetes.io/instance=hostpath.csi.k8s.io,app.kubernetes.io/part-of=csi-driver-host-path --wait=true
+kubectl delete role,clusterrole,rolebinding,clusterrolebinding,serviceaccount,storageclass,csidriver --all-namespaces -l app.kubernetes.io/instance=hostpath.csi.k8s.io,app.kubernetes.io/part-of=csi-driver-host-path --wait=true
--- a/deploy/util/deploy-hostpath.sh
+++ b/deploy/util/deploy-hostpath.sh
@@ -27,6 +27,10 @@ set -o pipefail
 
 BASE_DIR="$( cd "$( dirname "$0" )" && pwd )"
 
+TEMP_DIR="$( mktemp -d )"
+trap 'rm -rf ${TEMP_DIR}' EXIT
+
+
 # If set, the following env variables override image registry and/or tag for each of the images.
 # They are named after the image name, with hyphen replaced by underscore and in upper case.
 #
@@ -164,7 +168,28 @@ for component in CSI_PROVISIONER CSI_ATTACHER CSI_SNAPSHOTTER CSI_RESIZER CSI_EX
         echo "Using non-default RBAC rules for $component. Changes from $original to $current are:"
         diff -c <(wget --quiet -O - "$original") <(if [[ "$current" =~ ^http ]]; then wget --quiet -O - "$current"; else cat "$current"; fi) || true
     fi
-    run kubectl apply -f "${current}"
+
+    # using kustomize kubectl plugin to add labels to he rbac files.
+    # since we are deploying rbas directly with the url, the kustomize plugin only works with the local files
+    # we need to add the files locally in temp folder and using kustomize adding labels it will be applied
+    if [[ "${current}" =~ ^http:// ]] || [[ "${current}" =~ ^https:// ]]; then
+      run curl "${current}" --output "${TEMP_DIR}"/rbac.yaml --silent --location
+      current=./rbac.yaml
+    fi
+
+    cat <<- EOF > "${TEMP_DIR}"/kustomization.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+commonLabels:
+  app.kubernetes.io/instance: hostpath.csi.k8s.io
+  app.kubernetes.io/part-of: csi-driver-host-path
+
+resources:
+- ${current}
+EOF
+
+    run kubectl apply --kustomize "${TEMP_DIR}"
 done
 
 # deploy hostpath plugin and registrar sidecar
--- a/deploy/util/destroy-hostpath.sh
+++ b/deploy/util/destroy-hostpath.sh
@@ -1,0 +1,34 @@
+#!/usr/bin/env bash
+# Copyright 2021 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This script performs the step to destroy the hostpath plugin driver
+# completely by checking all the resources.
+# This should be in sync with the deploy script based on any future additions/deletions
+# of the resources
+
+# The script assumes that kubectl is available on the OS path
+# where it is executed.
+
+set -e
+set -o pipefail
+
+# Deleting all the resources installed by the deploy-script as the part of csi-hostpath-driver.
+# Every resource in the driver installation has the label representing the installation instance.
+# Using app.kubernetes.io/instance: hostpath.csi.k8s.io and app.kubernetes.io/part-of: csi-driver-host-path labels to identify the installation set
+#
+# kubectl maintains a few standard resources under "all" category which can be deleted by using just "kubectl delete all"
+# and other resources such as roles, clusterrole, serivceaccount etc needs to be deleted explicitly
+kubectl delete all --all-namespaces -l app.kubernetes.io/instance=hostpath.csi.k8s.io,app.kubernetes.io/part-of=csi-driver-host-path --wait=true
+kubectl delete role,clusterrole,rolebinding,clusterrolebinding,serviceaccount,storageclass,csidriver --all-namespaces -l app.kubernetes.io/instance=hostpath.csi.k8s.io,app.kubernetes.io/part-of=csi-driver-host-path --wait=true
--- a/deploy/kubernetes-1.18/destroy.sh
+++ b/deploy/kubernetes-1.18/destroy.sh
@@ -1,0 +1,1 @@
+../util/destroy-hostpath.sh
--- a/deploy/kubernetes-1.18/hostpath/csi-hostpath-attacher.yaml
+++ b/deploy/kubernetes-1.18/hostpath/csi-hostpath-attacher.yaml
@@ -2,23 +2,34 @@ kind: StatefulSet
 apiVersion: apps/v1
 metadata:
   name: csi-hostpath-attacher
+  labels:
+    app.kubernetes.io/instance: hostpath.csi.k8s.io
+    app.kubernetes.io/part-of: csi-driver-host-path
+    app.kubernetes.io/name: csi-hostpath-attacher
+    app.kubernetes.io/component: attacher
 spec:
   serviceName: "csi-hostpath-attacher"
   replicas: 1
   selector:
     matchLabels:
-      app: csi-hostpath-attacher
+      app.kubernetes.io/instance: hostpath.csi.k8s.io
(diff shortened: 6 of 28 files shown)


deploy/util/destroy-hostpath.sh
```diff
@@ -0,0 +1,56 @@
...
+
+# The script assumes that kubectl is available on the OS path
+# where it is executed.
+
+set -e
+set -o pipefail
+
+BASE_DIR=$(dirname "$0")
+
+LABELS=(csi-snapshotter csi-hostpathplugin csi-attacher csi-resizer csi-hostpath-driver csi-provisioner csi-hostpath-socat)
```
pohly, line 16: The labels have to be specific to the hostpath deployment. We cannot risk deleting something that may be part of some other CSI driver deployment.

The setup script has to add those labels instead of assuming that the sidecar RBAC objects already have certain labels. When the script was originally written, kustomize was still new. It might be possible to rely on the kustomize in kubectl >= 1.17 to add labels. Even the image selection might now be done that way.

pohly (changes requested):

deploy/util/destroy-hostpath.sh
aayushrangwala: Right. Makes more sense. I believe using kustomize across the project is in the scope of another ticket. But I will note your point and change the labels and maybe create a new pr for the kustomize

.gitignore
```diff
@@ -1 +1,17 @@
...
+.idea
 /bin
+
+# Test binary, built with `go test -c`
+*.test
+
+# Output of the go coverage tool, specifically when used with LiteIDE
+*.out
+
+# Dependency directories (remove the comment below to include it)
```
pohly, line 16: In your next push, please also clean up your commit:
- don't change .gitignore
- use a commit message that has a subject line and a body (separated by blank line) which explains what this is about and why the change is relevant

pohly: @aayushrangwala do you want to finish this PR?

aayushrangwala: @pohly Yeah, I will wrap this up. Got stuck in something completely unaboidable

aayushrangwala: cleaned commit

deploy/util/destroy-hostpath.sh
```diff
@@ -0,0 +1,56 @@
...
+
+# The script assumes that kubectl is available on the OS path
+# where it is executed.
+
+set -e
+set -o pipefail
+
+BASE_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
+LABELS=(csi-hostpath-snapshotter csi-hostpathplugin csi-hostpath-attacher csi-hostpath-resizer csi-hostpath-driver csi-hostpath-provisioner csi-hostpath-socat)
```
pohly, line 16: These labels are both not specific enough (are they guaranteed to be unique?!) and too complex. Consider https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/.

What we should be using is:
```
app.kubernetes.io/instance: hostpath.csi.k8s.io
app.kubernetes.io/part-of: csi-driver-host-path
```

Here `hostpath.csi.k8s.io` is the CSI driver name which is guaranteed to be unique in the cluster.

There are others that might make sense (like `app.kubernetes.io/name` and `app.kubernetes.io/component` to describe individual parts, e.g. `name: external-provisioner`, `component: provisioner`) but we don't need them for cleaning up.

With those two labels, we don't need to iterate over multiple labels when deleting.

deploy/util/destroy-hostpath.sh
```diff
@@ -0,0 +1,56 @@
...
+    done
+     
+}
+
+run
+
+# Wait until all pods are not deleted. We have to make some assumptions
+# about the deployment here, otherwise we wouldn't know what to wait
+# for: the expectation is that we run attacher, provisioner,
+# snapshotter, resizer, socat and hostpath plugin in the default namespace.
```
pohly, line 40: Instead of this logic here, you can add deleting of pods and use `kubectl delete --wait=true` for everything.

pohly: Throw in `--cascade=foreground` and deleting a daemonset will wait for pods to be deleted. Deleting the pods then becomes a safeguard that shouldn't be needed.

deploy/util/destroy-hostpath.sh
```diff
@@ -0,0 +1,56 @@
...
+    for label in "${LABELS[@]}"; do
+        kubectl delete all,role,clusterrole,rolebinding,clusterrolebinding,sa -l app=${label} || true
+    done
+
+    # delete hostpath plugin as a fall back
+    echo "deleting hostpath components"
+    for i in $(ls ${BASE_DIR}/hostpath/*.yaml | sort); do
+        echo "deleting $i"
+        kubectl delete -f "$i" || true
+    done
```
pohly, line 31: I don't think we need this.

deploy/util/destroy-hostpath.sh
aayushrangwala: Great!. Its way more suitable. made that change

deploy/util/destroy-hostpath.sh
aayushrangwala: Added that as a fallback. Can remove it

pohly: Please do. If deleting by label somehow is incomplete, we should fix that instead of pampering over the problem with a fallback.

aayushrangwala: Also adding the logic to delete rbacs. Since we install it using external (version calculated) files, we might have to delete them the same way as they dont have the required labels. Otherwise we have to add PRs on independent sidecars to add labels on all of their resources.

pohly: We can add labels when creating the RBAC objects. It makes the deploy script a bit more complicated, but the result then is consistent.

Using kustomize instead of the hand-crafted shell script is also an option.

deploy/util/destroy-hostpath.sh
aayushrangwala: Added the change

deploy/util/destroy-hostpath.sh
aayushrangwala: @pohly Added the labels to rbac

deploy/util/destroy-hostpath.sh
aayushrangwala: Resolved them. Please re-review

deploy/util/destroy-hostpath.sh
pohly: The more I look at this less convinced I am that we should have the code for deleting pods.

Grepping by name is going to fail if in the (unlikely, but not impossible) scenario that there are multiple deployments and we simply shouldn't need this.

deploy/util/destroy-hostpath.sh
```diff
@@ -0,0 +1,54 @@
...
+# The script assumes that kubectl is available on the OS path
+# where it is executed.
+
+set -e
+set -o pipefail
+
+BASE_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
+kubectl delete sts,ds,deploy,role,clusterrole,rolebinding,clusterrolebinding,sa -l app.kubernetes.io/instance=hostpath.csi.k8s.io,app.kubernetes.io/part-of=csi-driver-host-path --wait=true --timeout=60s --cascade=foreground
+kubectl delete all -l app.kubernetes.io/instance=hostpath.csi.k8s.io,app.kubernetes.io/part-of=csi-driver-host-path --wait=true --timeout=60s --cascade=foreground
```
pohly, line 30: Please add `--all-namespaces` to both commands.

Can you also spell out the object types instead of using the shorthands? This is going to be read more often than being written, and at least I had to stop and think about what "sa" and "sts" stand for.

pohly: Please add a comment in the script about why some objects get deleted explicitly and others not.

deploy/util/destroy-hostpath.sh
```diff
@@ -0,0 +1,54 @@
...
+# This should be in sync with the deploy script based on any future additions/deletions
+# of the resources
+
+# The script assumes that kubectl is available on the OS path
+# where it is executed.
+
+set -e
+set -o pipefail
+
+BASE_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
```
pohly, line 27: No need to depend on bash here.
Please use the same stanza as in `deploy-hostpath.sh`:
```
BASE_DIR="$( cd "$( dirname "$0" )" && pwd )"
```

pohly (changes requested):

deploy/util/destroy-hostpath.sh
aayushrangwala: Not needed because I removed the pod deletion logic

deploy/util/destroy-hostpath.sh
aayushrangwala: Added `--all-namespaces` and the comment. Also removed the deletion of pods based on grep, although the same logic is there in the deploy script to check if the pods are not up for each component

deploy/util/deploy-hostpath.sh
```diff
@@ -164,7 +164,21 @@ for component in CSI_PROVISIONER CSI_ATTACHER CSI_SNAPSHOTTER CSI_RESIZER CSI_EX
         echo "Using non-default RBAC rules for $component. Changes from $original to $current are:"
         diff -c <(wget --quiet -O - "$original") <(if [[ "$current" =~ ^http ]]; then wget --quiet -O - "$current"; else cat "$current"; fi) || true
     fi
-    run kubectl apply -f "${current}"
+
+    echo "adding labels to rbac"
+    cat <<- EOF > ./kustomization.yaml
```
pohly, line 169: We don't know where this script runs and shouldn't blindly overwrite the kustomization.yaml file there.

Please use `mktemp -d` to create a temporary directory at the start of this script with an exit trap to clean it up.

aayushrangwala: made the changes. Please check

deploy/util/deploy-hostpath.sh
```diff
@@ -27,6 +27,10 @@ set -o pipefail
 
 BASE_DIR="$( cd "$( dirname "$0" )" && pwd )"
 
+echo "creating temp directory"
```
pohly, line 30: I don't think we need to print this.

deploy/util/deploy-hostpath.sh
```diff
@@ -214,6 +234,9 @@ for i in $(ls ${BASE_DIR}/hostpath/*.yaml | sort); do
     fi
 done
 
+echo "deleting temp directory: ${TEMP_DIR}"
+rm -rf "${TEMP_DIR}"
```
pohly, line 238: A better way to do this is via an exit trap. Then the cleanup happens also when aborting with CTRL-C.

See above.

deploy/util/deploy-hostpath.sh
```diff
@@ -27,6 +27,10 @@ set -o pipefail
 
 BASE_DIR="$( cd "$( dirname "$0" )" && pwd )"
 
+echo "creating temp directory"
+TEMP_DIR="$( mktemp -d )"
```
pohly, line 31: Add here:
```
trap 'rm -r $TEMP_DIR' EXIT
```

pohly (changes requested):

deploy/util/deploy-hostpath.sh
```diff
@@ -164,7 +168,23 @@ for component in CSI_PROVISIONER CSI_ATTACHER CSI_SNAPSHOTTER CSI_RESIZER CSI_EX
...
+
+commonLabels:
+  app.kubernetes.io/instance: hostpath.csi.k8s.io
+  app.kubernetes.io/part-of: csi-driver-host-path
+
+resources:
+- ${current}
+EOF
+
+    curl "${current}" --output rbac.yaml
```
pohly, line 185: `${current}` might be a file when used for testing the sidecar repo which contains the RBAC file.

 `kubectl` handles that, curl doesn't. You need to add a check for that here and only download if it is a URL.

Please also add `--silent --location` to the curl invocation.

deploy/util/deploy-hostpath.sh
```diff
@@ -164,7 +168,23 @@ for component in CSI_PROVISIONER CSI_ATTACHER CSI_SNAPSHOTTER CSI_RESIZER CSI_EX
...
+    run kubectl apply --kustomize "${TEMP_DIR}"
+    run rm "${TEMP_DIR}"/kustomization.yaml
```
pohly, line 187: You can remove this `rm`.

pohly (changes requested):

deploy/util/deploy-hostpath.sh
aayushrangwala: got it

deploy/util/deploy-hostpath.sh
aayushrangwala: right, skipped my mind. Added

deploy/util/deploy-hostpath.sh
aayushrangwala: Good point. Added

deploy/util/deploy-hostpath.sh
aayushrangwala: got it. removed

deploy/util/deploy-hostpath.sh
aayushrangwala: done

deploy/util/deploy-hostpath.sh
```diff
@@ -164,7 +168,25 @@ for component in CSI_PROVISIONER CSI_ATTACHER CSI_SNAPSHOTTER CSI_RESIZER CSI_EX
...
+  app.kubernetes.io/instance: hostpath.csi.k8s.io
+  app.kubernetes.io/part-of: csi-driver-host-path
+
+resources:
+- ${current}
+EOF
+
+    if [[ "${current}" =~ ^http:// ]] || [[ "${current}" =~ ^https:// ]]; then
+      curl "${current}" --output rbac.yaml --silent --location
+    fi
```
pohly, line 187: ... and this `rbac.yaml` gets used where?

pohly (changes requested):

deploy/util/deploy-hostpath.sh
aayushrangwala: My bad. Skipped one commit to push. Will add the comment as well and push again

deploy/kubernetes-1.20/hostpath/csi-hostpath-plugin.yaml
```diff
@@ -19,6 +22,11 @@ kind: StatefulSet
 apiVersion: apps/v1
 metadata:
   name: csi-hostpathplugin
+  labels:
+    app.kubernetes.io/instance: hostpath.csi.k8s.io
+    app.kubernetes.io/part-of: csi-driver-host-path
+    app.kubernetes.io/name: csi-hostpathplugin
+    app.kubernetes.io/component: plugin
```
pohly, line 29: Those labels also should be set inside the PodSpec, otherwise `kubectl get all -l     app.kubernetes.io/instance=hostpath.csi.k8s.io` returns an incomplete result.


I think it is okay to use the same key/value pairs as for the StatefulSet itself.

The same change is needed also for the other StatefulSets and DaemonSets.

deploy/util/deploy-hostpath.sh
```diff
@@ -164,7 +168,30 @@ for component in CSI_PROVISIONER CSI_ATTACHER CSI_SNAPSHOTTER CSI_RESIZER CSI_EX
         echo "Using non-default RBAC rules for $component. Changes from $original to $current are:"
         diff -c <(wget --quiet -O - "$original") <(if [[ "$current" =~ ^http ]]; then wget --quiet -O - "$current"; else cat "$current"; fi) || true
     fi
-    run kubectl apply -f "${current}"
+
+    echo "adding labels to rbac"
```
pohly, line 172: I ran this script and did not fine this line useful. Can you remove it?

deploy/util/deploy-hostpath.sh
```diff
@@ -164,7 +168,30 @@ for component in CSI_PROVISIONER CSI_ATTACHER CSI_SNAPSHOTTER CSI_RESIZER CSI_EX
...
+
+    # using kustomize kubectl plugin to add labels to he rbac files.
+    # since we are deploying rbas directly with the url, the kustomize plugin only works with the local files
+    # we need to add the files locally in temp folder and using kustomize adding labels it will be applied
+    if [[ "${current}" =~ ^http:// ]] || [[ "${current}" =~ ^https:// ]]; then
+      curl "${current}" --output "${TEMP_DIR}"/rbac.yaml --silent --location
```
pohly, line 178: More important is to wrap this with `run`, because then it will be obvious what failed if curl prints error messages.

pohly (changes requested):
Can you also update deploy/kubernetes-distributed?

The deploy.sh script there is a fork, but your changes apply with some offsets:
```
$ git diff origin/master..HEAD deploy/util/deploy-hostpath.sh | patch deploy/kubernetes-distributed/deploy.sh
patching file deploy/kubernetes-distributed/deploy.sh
Hunk #1 succeeded at 13 with fuzz 2 (offset -14 lines).
Hunk #2 succeeded at 137 with fuzz 1 (offset -31 lines).
```

deploy/kubernetes-1.20/hostpath/csi-hostpath-plugin.yaml
aayushrangwala: Yeah makes sense it will make the child pods for the app object distinguishable

deploy/util/deploy-hostpath.sh
aayushrangwala: removed

deploy/util/deploy-hostpath.sh
aayushrangwala: right.

deploy/kubernetes-distributed/destroy.sh
```diff
@@ -0,0 +1,34 @@
...
+set -o pipefail
+
+# Deleting all the resources installed by the deploy-script as the part of csi-hostpath-driver.
+# Every resource in the driver installation has the label representing the installation instance.
+# Using app.kubernetes.io/instance: hostpath.csi.k8s.io and app.kubernetes.io/part-of: csi-driver-host-path labels to identify the installation set
+#
+# kubectl maintains a few standard resources under "all" category which can be deleted by using just "kubectl delete all"
+# and other resources such as roles, clusterrole, serivceaccount etc needs to be deleted explicitly
+kubectl delete all --all-namespaces -l app.kubernetes.io/instance=hostpath.csi.k8s.io,app.kubernetes.io/part-of=csi-driver-host-path --wait=true
+kubectl delete role,clusterrole,rolebinding,clusterrolebinding,serviceaccount,storageclass --all-namespaces -l app.kubernetes.io/instance=hostpath.csi.k8s.io,app.kubernetes.io/part-of=csi-driver-host-path --wait=true
```
pohly, line 34: `csidriver` must be added here.

pohly (changes requested):
One last fix and then it should be ready.

Please also squash all commits.

deploy/kubernetes-distributed/destroy.sh
aayushrangwala: Done

pohly (approved):
/lgtm
/approve

Thanks for the contribution.

Issue #199, opened 2020-09-14
k8s 1.17: csi-hostpath-snapclass is not created during the driver deployment


sedovalx:
I'm trying to deploy the driver on minikube with `--kubernetes-version v1.17.9`, so I have installed CRDs and then have run `csi-driver-host-path/deploy/kubernetes-1.17/deploy.sh` for that. The last line in the deploy log was "deploying snapshotclass based on snapshotter version". And then, when I tried to create a snapshot with `examples/csi-snapshot-v1beta1.yaml` I've got the following error for the `new-snapshot-demo` volume snapshot:
```
$ kubectl describe volumesnapshot                                                                
Name:         new-snapshot-demo
Namespace:    default
Labels:       <none>
Annotations:  kubectl.kubernetes.io/last-applied-configuration:
                {"apiVersion":"snapshot.storage.k8s.io/v1beta1","kind":"VolumeSnapshot","metadata":{"annotations":{},"name":"new-snapshot-demo","namespace...
API Version:  snapshot.storage.k8s.io/v1beta1
Kind:         VolumeSnapshot
Metadata:
  Creation Timestamp:  2020-09-14T10:21:34Z
  Generation:          1
  Resource Version:    59736
  Self Link:           /apis/snapshot.storage.k8s.io/v1beta1/namespaces/default/volumesnapshots/new-snapshot-demo
  UID:                 1acdcca0-2cbd-4825-b15d-1eb6edd793b8
Spec:
  Source:
    Persistent Volume Claim Name:  csi-pvc
  Volume Snapshot Class Name:      csi-hostpath-snapclass
Status:
  Error:
    Message:     Failed to get snapshot class with error failed to retrieve snapshot class csi-hostpath-snapclass from the informer: "volumesnapshotclass.snapshot.storage.k8s.io \"csi-hostpath-snapclass\" not found"
    Time:        2020-09-14T10:21:34Z
  Ready To Use:  false
Events:
  Type     Reason                  Age    From                 Message
  ----     ------                  ----   ----                 -------
  Warning  GetSnapshotClassFailed  4m23s  snapshot-controller  Failed to get snapshot class with error failed to retrieve snapshot class csi-hostpath-snapclass from the informer: "volumesnapshotclass.snapshot.storage.k8s.io \"csi-hostpath-snapclass\" not found"
```

[...]

PR #218, opened 2020-11-10
added the fix for script for not being able to use in script source path

aayushrangwala:
**What type of PR is this?**
/kind bug

The csi driver deployment script was using the base path as the curr dir where the script is triggerred, csi-snapshot class was not getting installed because of incorrect path resolution. Fixed to use the base path always using pwd and cd

Fixes #https://github.com/kubernetes-csi/csi-driver-host-path/issues/199

NONE

```release-note
Added the Fix for path resolution in deploy script. Now the user can run script from anywhere, not necessarily from the base path as `./deploy/kubernetes-1.1x/depploy-script.sh`
```

### DIFF
--- a/deploy/util/deploy-hostpath.sh
+++ b/deploy/util/deploy-hostpath.sh
@@ -11,7 +11,7 @@
 set -e
 set -o pipefail
 
-BASE_DIR=$(dirname "$0")
+BASE_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 
 # If set, the following env variables override image registry and/or tag for each of the images.
 # They are named after the image name, with hyphen replaced by underscore and in upper case.
--- a/deploy/util/deploy-hostpath.sh
+++ b/deploy/util/deploy-hostpath.sh
@@ -11,7 +11,7 @@
 set -e
 set -o pipefail
 
-BASE_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+BASE_DIR="$( cd "$( dirname "$0" )" && pwd )"
 
 # If set, the following env variables override image registry and/or tag for each of the images.
 # They are named after the image name, with hyphen replaced by underscore and in upper case.


deploy/util/deploy-hostpath.sh
```diff
@@ -11,7 +11,7 @@
 set -e
 set -o pipefail
 
-BASE_DIR=$(dirname "$0")
+BASE_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
```
pohly, line 14: Do we need `BASH_SOURCE` here? Why not just `$0`?

Also, why the output redirection?

aayushrangwala: Well both `BASH_SOURCE` and `$0` would serve the same purpose here just using the `BASH_SOURCE` in bash script is better even if it is sourced

And for output redirection, sometimes cd prints something to stdout, in the case where a CDPATH is set and even an error can be printed. So to be safe, there is the stdout redirection to dev null and stderr to stdout

To cover CDPATH weird cases, we can even follow to `unset CDPATH` at the begining and not use this redirection but still it will be changing user's config

pohly: > Well both BASH_SOURCE and $0 would serve the same purpose here just using the BASH_SOURCE in bash script is better even if it is sourced

I think it is better to not depend on bash features unnecessarily, because then a conversion to a non-Bash script becomes easier. Currently `pipefail` is the only bashism that I am aware of, and even that may not be necessary.

> And for output redirection, sometimes cd prints something to stdout

During normal operation? That's very unexpected.

> in the case where a CDPATH is set and even an error can be printed.

If an error is printed, presumably the return code is non-zero, `BASE_DIR` is unset and everything fails. We *want* to see the error message of `cd` in that case.

aayushrangwala: ohkay, right. we can depend on pipefail I can ignore the output redirection

aayushrangwala: done

pohly (approved):
/lgtm

pohly:
/assign @xing-yang 

For approval.

PR #221, opened 2020-11-12
WIP: Update snapshot CRD version

xing-yang:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note

```

### DIFF
--- a/release-tools/prow.sh
+++ b/release-tools/prow.sh
@@ -330,7 +330,14 @@ configvar CSI_PROW_E2E_ALPHA_GATES_LATEST 'GenericEphemeralVolume=true,CSIStorag
 configvar CSI_PROW_E2E_ALPHA_GATES "$(get_versioned_variable CSI_PROW_E2E_ALPHA_GATES "${csi_prow_kubernetes_version_suffix}")" "alpha E2E feature gates"
 
 # Which external-snapshotter tag to use for the snapshotter CRD and snapshot-controller deployment
-configvar CSI_SNAPSHOTTER_VERSION 'v3.0.0' "external-snapshotter version tag"
+default_csi_snapshotter_version () {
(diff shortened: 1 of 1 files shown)


release-tools/prow.sh
```diff
@@ -693,6 +693,10 @@ install_csi_driver () {
 install_snapshot_crds() {
   # Wait until volumesnapshot CRDs are in place.
   CRD_BASE_DIR="https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/${CSI_SNAPSHOTTER_VERSION}/client/config/crd"
+  if [ "${CSI_PROW_KUBERNETES_VERSION}" = "latest" ] || [ "${CSI_PROW_DRIVER_CANARY}" == "canary" ]; then
+	CRD_BASE_DIR="https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/master/client/config/crd"
+  fi
```
pohly, line 698: What about the controller? When running an "alpha" job, it probably makes more sense to also use the "canary" version of that.

I would change the default value of `CSI_SNAPSHOTTER_VERSION`:
```
default_csi_snapshotter_version () {
     if [ "${CSI_PROW_KUBERNETES_VERSION}" = "latest" ] || [ "${CSI_PROW_DRIVER_CANARY}" = "canary" ]; then
      echo "master"
     else
       echo "v3.0.0"
     fi
}
configvar CSI_SNAPSHOTTER_VERSION "$(default_csi_snapshotter_version)" "external-snapshotter version tag"
```

The advantage is that the prow job output correctly shows what's being used, instead of using something else under the hood.

pohly (changes requested):

release-tools/prow.sh
xing-yang: The controller is not updated yet.  sure, I'll make the suggested changes.

xing-yang:
Note: unit test will not pass on this PR because I modified release-tools manually.

PR #232, opened 2020-12-11
Fix deploy test-driver SizeRange and add logs

Jiawei0227:
**What type of PR is this?**
/kind cleanup

This PR does the following:
1. Add MinSizeRange for hostpath test-driver so the sig-storage test could pass
2. Add some logging for triage in the future
3. Add doc about the restore process time

This fix is mainly to fix a sig-storage test: https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-csi-1-19-on-kubernetes-master/1337029920152358912

The bug is captured because we add the defaultSizeRange in the external driver parsing. And that can lead to very slow block snapshot creation and leads to test failure.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/hostpath/server.go
+++ b/pkg/hostpath/server.go
@@ -114,6 +114,9 @@ func parseEndpoint(ep string) (string, string, error) {
 }
 
 func logGRPC(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+	if info.FullMethod == "/csi.v1.Identity/Probe" {
+		return handler(ctx, req)
+	}
 	glog.V(3).Infof("GRPC call: %s", info.FullMethod)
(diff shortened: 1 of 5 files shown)


pkg/hostpath/server.go
```diff
@@ -114,6 +114,9 @@ func parseEndpoint(ep string) (string, string, error) {
 }
 
 func logGRPC(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+	if info.FullMethod == "/csi.v1.Identity/Probe" {
```
msau42, line 117: is this needed?

Jiawei0227: This is not necessarily needed. But I dont see why we want to see continues probe spam our log. It is A LOT OF probe logs if we have livenessprobe. And I dont see the need to print them out

Issue #243, opened 2021-02-02
one snapshot create many snap file & much slow


daresheep:
Dears,

csi-driver-host-path version : V1.4 /V1.5
external-snapshotter:  v4.0

```
kubectl get pvc           
NAME            STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS      AGE
iso-win         Bound    pvc-d9de552d-0e49-4abd-a368-f22b9c92010b   28Gi       RWO            csi-hostpath-sc   38m
iso-win-block   Bound    pvc-16e7d86f-fd29-4bb9-b5cd-0f4c37af2d9d   28Gi       RWO            csi-hostpath-sc   19m
```

try to create a snapshot for iso-win use below yaml:

```
apiVersion: snapshot.storage.k8s.io/v1beta1
kind: VolumeSnapshot
metadata:
  name: new-snapshot-demo
spec:
  volumeSnapshotClassName: csi-hostpath-snapclass
  source:
    persistentVolumeClaimName: iso-win
```
```
kubectl apply -f csi-snapshot-v1beta1.yaml 
volumesnapshot.snapshot.storage.k8s.io/new-snapshot-demo created
```
watch at  csi-data-path ,a lot of snap files created. each 1min .

![image](https://user-images.githubusercontent.com/8828158/106571352-033df000-6572-11eb-98fb-5aa31bd45a77.png)

![image](https://user-images.githubusercontent.com/8828158/106685044-8eff5d00-6602-11eb-9bbe-10af44092963.png)

Here is some logs for  hostpath

```
kubectl logs -f csi-hostpathplugin-0 hostpath

I0203 11:12:21.879787       1 identityserver.go:40] Using default GetPluginInfo
I0203 11:12:21.879795       1 server.go:123] GRPC response: {"name":"hostpath.csi.k8s.io","vendor_version":"v1.4.0-rc2-0-gee6beeaf"}
I0203 11:12:21.883128       1 server.go:117] GRPC call: /csi.v1.Controller/CreateSnapshot
I0203 11:12:21.883142       1 server.go:118] GRPC request: {"name":"snapshot-023764b9-7119-4acf-8cd7-33ba0712c5cf","source_volume_id":"e5b4c966-656e-11eb-894d-5e012443d7b4"}
I0203 11:12:21.883640       1 controllerserver.go:337] Creating snapshot of Raw Block Mode Volume


[...]

PR #245, opened 2021-02-04
fix 1 snapshot create a lot of snap file

daresheep:
If the file task more then 1 min.  other request will create new snap file. 

issue:

https://github.com/kubernetes-csi/csi-driver-host-path/issues/243

### DIFF
--- a/pkg/hostpath/controllerserver.go
+++ b/pkg/hostpath/controllerserver.go
@@ -413,13 +413,8 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 		glog.V(4).Infof("Creating snapshot of Filsystem Mode Volume")
 		cmd = []string{"tar", "czf", file, "-C", volPath, "."}
 	}
-	executor := utilexec.New()
-	out, err := executor.Command(cmd[0], cmd[1:]...).CombinedOutput()
-	if err != nil {
-		return nil, status.Errorf(codes.Internal, "failed create snapshot: %v: %s", err, out)
-	}
 
-	glog.V(4).Infof("create volume snapshot %s", file)
+	//since other process need check current snapshot id
 	snapshot := hostPathSnapshot{}
 	snapshot.Name = req.GetName()
 	snapshot.Id = snapshotID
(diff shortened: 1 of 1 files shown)


daresheep:
CLA signature

pkg/hostpath/controllerserver.go
```diff
@@ -413,20 +413,26 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
...
+	//since other process need check current snapshot id
 	snapshot := hostPathSnapshot{}
 	snapshot.Name = req.GetName()
 	snapshot.Id = snapshotID
 	snapshot.VolID = volumeID
 	snapshot.Path = file
 	snapshot.CreationTime = creationTime
 	snapshot.SizeBytes = hostPathVolume.VolSize
+	snapshot.ReadyToUse = false
+	hostPathVolumeSnapshots[snapshotID] = snapshot
```
aayushrangwala, line 426: This entry is being added below as well `:438` which makes more sense there as anyway the snapshot_id is being added to map only after the successful creation.

pkg/hostpath/controllerserver.go
```diff
@@ -413,20 +413,26 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
...
-	glog.V(4).Infof("create volume snapshot %s", file)
+	//since other process need check current snapshot id
 	snapshot := hostPathSnapshot{}
 	snapshot.Name = req.GetName()
 	snapshot.Id = snapshotID
 	snapshot.VolID = volumeID
 	snapshot.Path = file
 	snapshot.CreationTime = creationTime
 	snapshot.SizeBytes = hostPathVolume.VolSize
+	snapshot.ReadyToUse = false
```
aayushrangwala, line 425: `snapshot.ReadyToUse` must be false by default as its a bool with zero value `false`. There seems to be no need here

pkg/hostpath/controllerserver.go
```diff
@@ -413,20 +413,26 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
...
+	hostPathVolumeSnapshots[snapshotID] = snapshot
+
+	executor := utilexec.New()
+	out, err := executor.Command(cmd[0], cmd[1:]...).CombinedOutput()
+	if err != nil {
+		delete(hostPathVolumeSnapshots,snapshotID)
+		return nil, status.Errorf(codes.Internal, "failed create snapshot: %v: %s", err, out)
+	}
+
+	glog.V(4).Infof("create volume snapshot %s", file)
```
aayushrangwala, line 435: log is incorrect. It should be `Craeted`

aayushrangwala:
This overall fix in the PR doesnot seems to make much difference. The issue seems to be there for the large PVs and its entry is not being added to the map and in the next iteration `getSnapshotByName()` may not be returning anything.

Need to check if the command is executed without error or is it going timeout

PR #248, opened 2021-02-12
Storage capacity

### LINKED ISSUE #72: map access needs lock

pohly:
**What type of PR is this?**
/kind feature

We need a way to demonstrate and test both distributed provisioning and storage capacity tracking.

This PR also fixes the long-standing issue of not properly protecting global variables against concurrent access.

Fixes #72

This needs to be complemented by special Prow jobs that use the new deployment: https://github.com/kubernetes/test-infra/pull/20848

```release-note
simulate storage capacity constraints and publish available capacity information
```

### DIFF
--- a/deploy/kubernetes-distributed/hostpath/csi-hostpath-plugin.yaml
+++ b/deploy/kubernetes-distributed/hostpath/csi-hostpath-plugin.yaml
@@ -1,0 +1,155 @@
+kind: DaemonSet
+apiVersion: apps/v1
+metadata:
+  name: csi-hostpathplugin
+spec:
+  selector:
+    matchLabels:
+      app: csi-hostpathplugin
+  template:
+    metadata:
+      labels:
+        app: csi-hostpathplugin
+    spec:
+      serviceAccount: csi-provisioner
+      containers:
+        - name: csi-provisioner
+          image: k8s.gcr.io/sig-storage/csi-provisioner:v2.1.0
+          args:
+            - -v=5
+            - --csi-address=/csi/csi.sock
+            - --feature-gates=Topology=true
+            - --enable-capacity
+            - --capacity-ownerref-level=0 # pod is owner
+            - --node-deployment=true
+            - --strict-topology=true
+            - --immediate-topology=false
+            - --worker-threads=5
+          env:
+          - name: NODE_NAME
+            valueFrom:
+              fieldRef:
+                apiVersion: v1
+                fieldPath: spec.nodeName
+          - name: POD_NAMESPACE
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.namespace
+          - name: POD_NAME
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.name
+          securityContext:
+            # This is necessary only for systems with SELinux, where
+            # non-privileged sidecar containers cannot access unix domain socket
+            # created by privileged CSI driver container.
+            privileged: true
+          volumeMounts:
+            - mountPath: /csi
+              name: socket-dir
+
+        - name: node-driver-registrar
+          image: k8s.gcr.io/sig-storage/csi-node-driver-registrar:v2.0.1
+          args:
+            - --v=5
+            - --csi-address=/csi/csi.sock
+            - --kubelet-registration-path=/var/lib/kubelet/plugins/csi-hostpath/csi.sock
+          securityContext:
+            # This is necessary only for systems with SELinux, where
+            # non-privileged sidecar containers cannot access unix domain socket
+            # created by privileged CSI driver container.
+            privileged: true
+          env:
+            - name: KUBE_NODE_NAME
+              valueFrom:
+                fieldRef:
+                  apiVersion: v1
+                  fieldPath: spec.nodeName
+          volumeMounts:
+          - mountPath: /csi
+            name: socket-dir
+          - mountPath: /registration
+            name: registration-dir
+          - mountPath: /csi-data-dir
+            name: csi-data-dir
+
+        - name: hostpath
+          image: k8s.gcr.io/sig-storage/hostpathplugin:v1.6.0
+          args:
+            - --drivername=hostpath.csi.k8s.io
+            - --v=5
+            - --endpoint=$(CSI_ENDPOINT)
+            - --nodeid=$(KUBE_NODE_NAME)
+            - --capacity=slow=10Gi
+            - --capacity=fast=100Gi
+          env:
+            - name: CSI_ENDPOINT
+              value: unix:///csi/csi.sock
+            - name: KUBE_NODE_NAME
+              valueFrom:
+                fieldRef:
+                  apiVersion: v1
+                  fieldPath: spec.nodeName
+          securityContext:
+            privileged: true
+          ports:
+          - containerPort: 9898
+            name: healthz
+            protocol: TCP
+          livenessProbe:
+            failureThreshold: 5
+            httpGet:
+              path: /healthz
+              port: healthz
+            initialDelaySeconds: 10
+            timeoutSeconds: 3
+            periodSeconds: 2
+          volumeMounts:
+            - mountPath: /csi
+              name: socket-dir
+            - mountPath: /var/lib/kubelet/pods
+              mountPropagation: Bidirectional
+              name: mountpoint-dir
+            - mountPath: /var/lib/kubelet/plugins
+              mountPropagation: Bidirectional
+              name: plugins-dir
+            - mountPath: /csi-data-dir
+              name: csi-data-dir
+            - mountPath: /dev
+              name: dev-dir
+        - name: liveness-probe
+          volumeMounts:
+          - mountPath: /csi
+            name: socket-dir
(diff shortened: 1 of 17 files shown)


pohly:
I can split out the locking change into a separate PR, if desired.

deploy/kubernetes-distributed/hostpath/csi-hostpath-plugin.yaml
```diff
@@ -0,0 +1,155 @@
...
+          volumeMounts:
+          - mountPath: /csi
+            name: socket-dir
+          - mountPath: /registration
+            name: registration-dir
+          - mountPath: /csi-data-dir
+            name: csi-data-dir
+
+        - name: hostpath
+          image: k8s.gcr.io/sig-storage/hostpathplugin:v1.6.0
```
pohly, line 77: Originally I had "canary" here with a TODO. I think we can switch directly to using the next release number, then we avoid some churn.

pkg/hostpath/hostpath.go
```diff
@@ -343,6 +375,10 @@ func (hp *hostPath) deleteVolume(volID string) error {
 	if err := os.RemoveAll(path); err != nil && !os.IsNotExist(err) {
 		return err
 	}
+	glog.V(4).Infof("deleting hostpath volume: %s = %+v", volID, vol)
```
verult, line 378: Should we log this before the `RemoveAll`?

pkg/hostpath/hostpath.go
```diff
@@ -255,9 +265,29 @@ func getVolumePath(volID string) string {
...
 }
 
-// createVolume create the directory for the hostpath volume.
-// It returns the volume path or err if one occurs.
-func createHostpathVolume(volID, name string, cap int64, volAccessType accessType, ephemeral bool) (*hostPathVolume, error) {
+// createVolume allocates capacity, creates the directory for the hostpath volume, and
+// adds the volume to the list.
+//
+// It returns the volume path or err if one occurs. That error is suitable as result of a gRPC call.
+func (hp *hostPath) createVolume(volID, name string, cap int64, volAccessType accessType, ephemeral bool, kind string) (hpv *hostPathVolume, finalErr error) {
```
verult, line 272: Should we return invalid arg error if kind is specified but capacity is disabled?

pkg/hostpath/hostpath.go
```diff
@@ -54,9 +55,12 @@ type hostPath struct {
 	ephemeral         bool
 	maxVolumesPerNode int64
 
-	ids *identityServer
-	ns  *nodeServer
-	cs  *controllerServer
+	// gRPC calls involving any of the fields below must be serialized
+	// by locking this mutex before starting. Internal helper
+	// functions assume that the mutex has been locked.
+	mutex     sync.Mutex
```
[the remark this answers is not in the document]
verult: And is it worth having a RWMutex instead?

verult (commented):
This is cool! Largely looks good to me

pkg/hostpath/hostpath.go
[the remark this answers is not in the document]
pohly: A production-quality driver would definitely want to be more intelligent about its locking. But this is just a demo driver where the main priority is correctness, not performance when executing many operations in parallel.
Therefore I prefer to keep the locking as simple as possible, i.e. lock in the gRPC functions after inspecting parameters and before using local state. The rule for helper functions is simple: helpers don't lock and a gRPC function must take the lock before calling them. Only a few read-only functions may (but don't have to) be treated as exceptions.

What you propose for example may have a correctness issues: consider `CreateVolume` function checks for a volume with `GetVolume`, then proceeds to create it. In parallel, the function is called again for the same volume and also proceeds to create it. Both complete and we leak one volume because only one of them gets stored and returned to the caller.

pkg/hostpath/hostpath.go
```diff
@@ -126,14 +125,22 @@ func NewHostPathDriver(driverName, nodeID, endpoint string, ephemeral bool, maxV
...
 		nodeID:            nodeID,
 		endpoint:          endpoint,
 		ephemeral:         ephemeral,
 		maxVolumesPerNode: maxVolumesPerNode,
-	}, nil
+
+		volumes:   map[string]hostPathVolume{},
+		snapshots: map[string]hostPathSnapshot{},
+	}
+	hp.discoverExistingSnapshots()
```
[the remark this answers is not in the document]
pohly: I don't think it matters and I didn't have any particular reason for changing it. I can revert the order, just to be safe.

pkg/hostpath/controllerserver.go
```diff
@@ -145,8 +140,8 @@ func (hp *hostPath) CreateVolume(ctx context.Context, req *csi.CreateVolumeReque
 	}
 
 	volumeID := uuid.NewUUID().String()
-
-	vol, err := createHostpathVolume(volumeID, req.GetName(), capacity, requestedAccessType, false /* ephemeral */)
+	kind := req.GetParameters()[storageKind]
+	vol, err := hp.createVolume(volumeID, req.GetName(), capacity, requestedAccessType, false /* ephemeral */, kind)
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "failed to create volume %v: %v", volumeID, err)
```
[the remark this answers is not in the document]
pohly: Good catch, thanks!

I think all of these cases where we wrap some error can simply be replaced with returning the error or wrapping it. The result will be that instead of the non-descriptive `codes.Internal` the caller will get a `codes.Unknown`, which has the same effect.

There's also an opportunity to remove duplicate code by letting the `getVolume/SnapshotBy...` functions return a `NotFound` error themselves.

pkg/hostpath/hostpath.go
pohly: Hmm, I put it after the work is done to mirror the "adding hostpath volume" message, i.e. once we get that message, we know that `hp.volumes` is really being changed.

pkg/hostpath/hostpath.go
pohly: Makes sense, added.

pkg/hostpath/hostpath.go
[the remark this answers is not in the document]
verult: Ah yeah thanks for the catch, sgtm

pkg/hostpath/nodeserver.go
```diff
@@ -175,9 +176,9 @@ func (hp *hostPath) NodePublishVolume(ctx context.Context, req *csi.NodePublishV
 		}
 	}
 
-	volume := hostPathVolumes[req.GetVolumeId()]
+	volume := hp.volumes[req.GetVolumeId()]
```
verult, line 179: Could the requested volume ID not exist?

pkg/hostpath/hostpath.go
verult: just a nit, but maybe we could move this after *all* the work is done (i.e. after `delete()`) and log "*added/deleted* hostpath volume"

pkg/hostpath/hostpath.go
```diff
@@ -428,7 +428,7 @@ func (hp *hostPath) loadFromSnapshot(size int64, snapshotId, destPath string, mo
 	out, err := executor.Command(cmd[0], cmd[1:]...).CombinedOutput()
 	glog.V(4).Infof("Command Finish: %v", string(out))
 	if err != nil {
-		return status.Errorf(codes.Internal, "failed pre-populate data from snapshot %v: %v: %s", snapshotId, err, out)
+		return fmt.Errorf("failed pre-populate data from snapshot %v: %w: %s", snapshotId, err, out)
```
verult, line 433: There's one more instance of `codes.Internal` in this function above. Should that be changed too?

verult (approved):
/lgtm
/approve

(I don't think I have approval permissions though)

pkg/hostpath/hostpath.go
pohly: I had kept it for consistency with the two error returns around it which both return a status error. But thinking about it again, I agree that it is better to also change it. Done.

pkg/hostpath/hostpath.go
pohly: Yes, that's better. Changed.

pkg/hostpath/nodeserver.go
pohly: Hmm, good question. This change came from a global seach/replace, I haven't really thought about it.

There was indeed a check that the volume exists earlier and we even have a copy of that volume in `vol`. So the code is correct, just a bit unusual. I think this is better:
```
-       volume := hp.volumes[req.GetVolumeId()]
-       volume.NodeID = hp.nodeID
-       hp.volumes[req.GetVolumeId()] = volume
+       vol.NodeID = hp.nodeID
+       hp.updateVolume(req.GetVolumeId(), vol)
        return &csi.NodePublishVolumeResponse{}, nil
```

pohly:
/assign @jsafrane 

For approval.

Issue #247, opened 2021-02-09
merge mock and hostpath driver


pohly:
Both mock and hostpath driver are used for testing, just in different configurations and with different capabilities of the driver. It's not clear whether there actually is a reason to maintain two different drivers. Perhaps they should be merged into one.

Existing capabilities:
- check credentials (mock)
- optional attach (mock)
- snapshots (both?)
- error injection via JavaScript hooks (mock - may become obsolete via https://github.com/kubernetes/kubernetes/pull/97069)
- raw block volumes (hostpath)
- actual storage of data, including snapshots (hostpath)
- simulating capacity (hostpath)
- multi-node deployment (hostpath)

Missing capabilities:
- support non-default kubelet path (https://github.com/kubernetes-csi/csi-driver-host-path/issues/71) and use that to avoid leaking loop devices (https://github.com/kubernetes-sigs/kind/issues/1248)

pohly:
We discussed this and came to the conclusion that as long as we only need to maintain one driver it doesn't matter too much which one we pick. However, the hostpath driver seems to be a better starting point because it is more widely used (several Prow jobs), is more realistic (actual mounts) and more complete (raw block support), and can be released independently because it is in its own repo. The mock driver currently must be released together with csi-sanity.

So here's a tentative plan:
- move functionality from the mock driver to hostpath:
  - JSON call logging (https://github.com/kubernetes-csi/csi-test/blob/c4bffc15d5e956e1dee5ba88d1c1e007960774c2/driver/driver.go#L216-L243)
  - proxy (https://github.com/kubernetes-csi/csi-test/commit/ef07bafd81bf1b460f29f5f4972736caa2594851)
[...]

PR #260, opened 2021-03-19
Copy features from Mock CSI driver

avalluri:
**What type of PR is this?**
/kind feature

This the first step towards the goal of  replacing the Mock CSI driver with host-path driver in Kubernetes E2E testing as described in #247 

Fixes #

```release-note
- Added 'proxy' support; where the host-path driver offloads the CSI requests transparently to the external CSI driver running at `--proxy-endpoint`.
- Implemented Controller{Publish,Unpublish}Volume calls. Introduced a new command-line argument `--enable-attach` (defaults to `false`) which controls if the driver should add `RPC_PUBLISH_UNPUBLISH_VOLUME` to its controller capablilities.
- GRPC calls are logged in a unified(request, reply, and error) JSON format.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,10 @@
 
 This repository hosts the CSI Hostpath driver and all of its build and dependent configuration files to deploy the driver.
 
+---
+*WARNING: This driver is just a demo implementation and is used for CI testing. This has many fake implementations and other non-standard best practices, and should not be used as an example of how to write a real driver.
+---
+
 ## Pre-requisite
 - Kubernetes cluster
 - Running version 1.13 or later
--- a/README.md
+++ b/README.md
@@ -12,6 +12,14 @@ This repository hosts the CSI Hostpath driver and all of its build and dependent
 - Access to terminal with `kubectl` installed
 - For Kubernetes 1.17 or later, the VolumeSnapshot beta CRDs and Snapshot Controller must be installed as part of the cluster deployment (see Kubernetes 1.17+ deployment instructions)
 
+## Features
+
+The driver can provide empty directories that are backed by the same filesystem as EmptyDir volumes. In addition, it can provide raw block volumes that are backed by a single file in that same filesystem and bound to a loop device.
+
+[Various command line parameters](cmd/hostpathplugin/maing.go) influence the behavior of the driver. This is relevant in particular for the end-to-end testing that this driver is used for in Kubernetes.
+
+Usually, the driver implements all CSI operations itself. When deployed with the `-proxy-endpoint` parameter, it instead proxies all incoming connections for a CSI driver that is [embedded inside the Kubernetes E2E test suite](https://github.com/kubernetes/kubernetes/tree/master/test/e2e/storage/drivers/csi-test) and used for mocking a CSI driver [with callbacks provided by certain tests](https://github.com/kubernetes/kubernetes/blob/5ad79eae2dcbf33df3b35c48ec993d30fbda46dd/test/e2e/storage/csi_mock_volume.go#L110).
+
 ## Deployment
 Deployment varies depending on the Kubernetes version your cluster is running:
 - [Deployment for Kubernetes 1.17 and later](docs/deploy-1.17-and-later.md)
--- a/cmd/hostpathplugin/main.go
+++ b/cmd/hostpathplugin/main.go
@@ -41,6 +41,7 @@ var (
 		flag.Var(c, "capacity", "Simulate storage capacity. The parameter is <kind>=<quantity> where <kind> is the value of a 'kind' storage class parameter and <quantity> is the total amount of bytes for that kind. The flag may be used multiple times to configure different kinds.")
 		return c
 	}()
+	enableAttach = flag.Bool("enable-attach", false, "Enables RPC_PUBLISH_UNPUBLISH_VOLUME capability.")
 	// Set by the build process
 	version = ""
 )
@@ -58,7 +59,7 @@ func main() {
 		fmt.Fprintln(os.Stderr, "Deprecation warning: The ephemeral flag is deprecated and should only be used when deploying on Kubernetes 1.15. It will be removed in the future.")
 	}
 
-	driver, err := hostpath.NewHostPathDriver(*driverName, *nodeID, *endpoint, *ephemeral, *maxVolumesPerNode, version, capacity)
+	driver, err := hostpath.NewHostPathDriver(*driverName, *nodeID, *endpoint, *ephemeral, *maxVolumesPerNode, version, capacity, *enableAttach)
 	if err != nil {
 		fmt.Printf("Failed to initialize driver: %s", err.Error())
 		os.Exit(1)
--- a/cmd/hostpathplugin/main.go
+++ b/cmd/hostpathplugin/main.go
@@ -17,11 +17,16 @@ limitations under the License.
 package main
 
 import (
+	"context"
 	"flag"
 	"fmt"
 	"os"
+	"os/signal"
 	"path"
+	"syscall"
 
+	"github.com/golang/glog"
+	"github.com/kubernetes-csi/csi-driver-host-path/internal/proxy"
 	"github.com/kubernetes-csi/csi-driver-host-path/pkg/hostpath"
 )
 
@@ -30,7 +35,7 @@ func init() {
 }
 
 var (
-	endpoint          = flag.String("endpoint", "unix://tmp/csi.sock", "CSI endpoint")
+	csiEndpoint       = flag.String("endpoint", "unix://tmp/csi.sock", "CSI endpoint")
 	driverName        = flag.String("drivername", "hostpath.csi.k8s.io", "name of the driver")
 	nodeID            = flag.String("nodeid", "", "node id")
 	ephemeral         = flag.Bool("ephemeral", false, "publish volumes in ephemeral mode even if kubelet did not ask for it (only needed for Kubernetes 1.15)")
@@ -42,6 +47,9 @@ var (
 		return c
 	}()
 	enableAttach = flag.Bool("enable-attach", false, "Enables RPC_PUBLISH_UNPUBLISH_VOLUME capability.")
+	// The proxy-endpoint option is intended to used by the Kubernetes E2E test suite
+	// for proxying incoming calls to the embedded mock CSI driver.
+	proxyEndpoint = flag.String("proxy-endpoint", "", "Instead of running the CSI driver code, just proxy connections from csiEndpoint to the given listening socket.")
 	// Set by the build process
 	version = ""
 )
@@ -59,7 +67,30 @@ func main() {
 		fmt.Fprintln(os.Stderr, "Deprecation warning: The ephemeral flag is deprecated and should only be used when deploying on Kubernetes 1.15. It will be removed in the future.")
 	}
 
-	driver, err := hostpath.NewHostPathDriver(*driverName, *nodeID, *endpoint, *ephemeral, *maxVolumesPerNode, version, capacity, *enableAttach)
+	if *proxyEndpoint != "" {
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+		closer, err := proxy.Run(ctx, *csiEndpoint, *proxyEndpoint)
+		if err != nil {
+			glog.Fatalf("failed to run proxy: %v", err)
+		}
+		defer closer.Close()
+
+		// Wait for signal
+		sigc := make(chan os.Signal, 1)
+		sigs := []os.Signal{
+			syscall.SIGTERM,
+			syscall.SIGHUP,
+			syscall.SIGINT,
+			syscall.SIGQUIT,
+		}
+		signal.Notify(sigc, sigs...)
+
+		<-sigc
+		return
+	}
+
+	driver, err := hostpath.NewHostPathDriver(*driverName, *nodeID, *csiEndpoint, *ephemeral, *maxVolumesPerNode, version, capacity, *enableAttach)
 	if err != nil {
 		fmt.Printf("Failed to initialize driver: %s", err.Error())
 		os.Exit(1)
--- a/pkg/hostpath/controllerserver.go
+++ b/pkg/hostpath/controllerserver.go
@@ -202,6 +202,17 @@ func (hp *hostPath) DeleteVolume(ctx context.Context, req *csi.DeleteVolumeReque
 	defer hp.mutex.Unlock()
 
 	volId := req.GetVolumeId()
+	vol, err := hp.getVolumeByID(volId)
+	if err != nil {
+		// Volume not found: might have already deleted
+		return &csi.DeleteVolumeResponse{}, nil
+	}
+
+	if vol.IsAttached || vol.IsPublished || vol.IsStaged {
+		return nil, status.Errorf(codes.Internal, "Volume '%s' is still used (attached: %v, staged: %v, published: %v) by '%s' node",
+			vol.VolID, vol.IsAttached, vol.IsStaged, vol.IsPublished, vol.NodeID)
+	}
+
 	if err := hp.deleteVolume(volId); err != nil {
 		return nil, fmt.Errorf("failed to delete volume %v: %w", volId, err)
 	}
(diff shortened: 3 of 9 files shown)


pkg/hostpath/server.go
```diff
@@ -114,16 +115,39 @@ func parseEndpoint(ep string) (string, string, error) {
 }
 
 func logGRPC(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
-	if info.FullMethod == "/csi.v1.Identity/Probe" {
-		return handler(ctx, req)
-	}
-	glog.V(3).Infof("GRPC call: %s", info.FullMethod)
-	glog.V(5).Infof("GRPC request: %+v", protosanitizer.StripSecrets(req))
```
pohly, line 121: Let's keep the behavior at level 3 as it is:
- don't log `Probe`
- log as plain text as before

At level 5, we then *add* (not *replace*):
- plain text logging of `Probe`
- JSON output

`if klog.V(5).Enabled() { ... do JSON logging ... }` is a useful pattern for avoiding the overhead of formatting as JSON.

go.sum
```diff
@@ -415,11 +414,9 @@ github.com/olekukonko/tablewriter v0.0.0-20170122224234-a0225b3f23b5/go.mod h1:v
 github.com/onsi/ginkgo v0.0.0-20170829012221-11459a886d9c/go.mod h1:lLunBs/Ym6LB5Z9jYTR76FiuTmxDTDusOGeTQH+WWjE=
 github.com/onsi/ginkgo v1.6.0/go.mod h1:lLunBs/Ym6LB5Z9jYTR76FiuTmxDTDusOGeTQH+WWjE=
 github.com/onsi/ginkgo v1.8.0/go.mod h1:lLunBs/Ym6LB5Z9jYTR76FiuTmxDTDusOGeTQH+WWjE=
-github.com/onsi/ginkgo v1.11.0 h1:JAKSXpt1YjtLA7YpPiqO9ss6sNXEsPfSGdwN0UHqzrw=
```
pohly, line 418: Do the go.sum changes really belong into this commit?

cmd/hostpathplugin/main.go
```diff
@@ -41,6 +41,7 @@ var (
 		flag.Var(c, "capacity", "Simulate storage capacity. The parameter is <kind>=<quantity> where <kind> is the value of a 'kind' storage class parameter and <quantity> is the total amount of bytes for that kind. The flag may be used multiple times to configure different kinds.")
 		return c
 	}()
+	disableAttach = flag.Bool("disable-attach", false, "Disables RPC_PUBLISH_UNPUBLISH_VOLUME capability.")
```
pohly, line 49: `-disable-attach=false` is a double negation, which is harder to understand. Also considering that attach is not really doing anything useful (i.e. it's just for testing), I prefer `enable-attach` with false as default.

pkg/hostpath/controllerserver.go
```diff
@@ -99,7 +99,7 @@ func (hp *hostPath) CreateVolume(ctx context.Context, req *csi.CreateVolumeReque
 	defer hp.mutex.Unlock()
 
 	capacity := int64(req.GetCapacityRange().GetRequiredBytes())
-	topologies := []*csi.Topology{&csi.Topology{
+	topologies := []*csi.Topology{{
```
pohly, line 102: A useful enhancement, but better put it into a separate commit.

Nit: I prefer
```
topologies := []*csi.Topology{
    {Segments: map[string]string{TopologyKeyNode: hp.nodeID}},
}
```

pkg/hostpath/controllerserver.go
```diff
@@ -254,11 +254,82 @@ func (hp *hostPath) ValidateVolumeCapabilities(ctx context.Context, req *csi.Val
...
+
+	v, err := hp.getVolumeByID(req.VolumeId)
+	if err != nil {
+		// Not an error: a non-existent volume is not published.
+		// See also https://github.com/kubernetes-csi/external-attacher/pull/165
+		return &csi.ControllerUnpublishVolumeResponse{}, nil
+	}
+
+	// Check to see if the volume is already unpublished.
+	v.VolAttached = false
```
pohly, line 329: There is no check here?

pkg/hostpath/controllerserver.go
```diff
@@ -254,11 +254,82 @@ func (hp *hostPath) ValidateVolumeCapabilities(ctx context.Context, req *csi.Val
...
+	return &csi.ControllerPublishVolumeResponse{
+		PublishContext: map[string]string{},
+	}, nil
 }
 
 func (hp *hostPath) ControllerUnpublishVolume(ctx context.Context, req *csi.ControllerUnpublishVolumeRequest) (*csi.ControllerUnpublishVolumeResponse, error) {
-	return nil, status.Error(codes.Unimplemented, "")
+	if hp.disableAttach {
+		return nil, status.Error(codes.Unimplemented, "ControllerUnpublishVolume is not supported")
+	}
```
pohly, line 307: Which checks are needed in the other method calls when attaching is enabled?

I suppose
- NodePublishVolume without prior VolumePublishVolume must fail
- VolumeUnpublishVolume between NodePublishVolume and NodeUnpublishVolume must fail
- DeleteVolume between ControllerPublishVolume and ControllerUnpublishVolume must fail

pkg/hostpath/controllerserver.go
```diff
@@ -254,11 +254,82 @@ func (hp *hostPath) ValidateVolumeCapabilities(ctx context.Context, req *csi.Val
...
+
+	hp.mutex.Lock()
+	defer hp.mutex.Unlock()
+
+	v, err := hp.getVolumeByID(req.VolumeId)
+	if err != nil {
+		// Not an error: a non-existent volume is not published.
+		// See also https://github.com/kubernetes-csi/external-attacher/pull/165
+		return &csi.ControllerUnpublishVolumeResponse{}, nil
+	}
```
pohly, line 326: Check after this whether NodePublishVolume has been called without NodeUnpublishVolume.

pohly (changes requested):

pkg/hostpath/server.go
avalluri: > Let's keep the behavior at level 3 as it is:
> 
> * don't log `Probe`
> * log as plain text as before

@pohly  You mean logging the method name, or logging full request in plain text?

Does this looks right?:
```
	resp, err := handler(ctx, req)
	if info.FullMethod == "/csi.v1.Identity/Probe" {
		glog.V(5).Infof("GRPC call: %s", info.FullMethod)
		return resp, err
	}
	glog.V(3).Infof("GRPC call: %s", info.FullMethod)
	if glog.V(5) {
		// Strip secrets from request before logging
		json.Unmarshal([]byte(protosanitizer.StripSecrets(req).String()), req)
		logGRPCJson(info.FullMethod, req, resp, err)
	}
```

pohly: I mean that at level 5, we log `Probe` like any other method call. We don't return early either.

```

func logGRPC(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
    pri := 3
    if info.FullMethod == "/csi.v1.Identity/Probe" {
                // This call occurs frequently, therefore it only gets log at level 5.
		pri = 5
    }
    glog.V(pri).Infof("GRPC call: %s", info.FullMethod)

    var strippedReq interface{}
    v5 := glog.V(5)
    if v5.Enabled() {
          strippedReq = protosanitizer.StripSecrets(req)
          v5.Infof("GRPC request: %+v", strippedReq)
    }
    resp, err := handler(ctx, req)
    if err != nil {
        // Always log errors. Probably not useful though without the method name?!
	glog.Errorf("GRPC error: %v", err)
    }
    if v5.Enabled() {
        strippedResp := protosanitizer.StripSecrets(resp)
	v5.Infof("GRPC response: %+v", strippedResp)
   
        logGRPCJson(info.FullMethod, strippedReq, strippedResp, err)
    }

    return resp, err
}
```

avalluri: @pohly I took your changes with minor fixes. Thanks.

go.sum
avalluri: This result of Go version update, which is not valid anymore.

cmd/hostpathplugin/main.go
avalluri: Agreed, and changed as suggested.

pkg/hostpath/controllerserver.go
avalluri: Moved to a separate commit.

pkg/hostpath/controllerserver.go
avalluri: Removed this unneeded comment, and added missing checks for validating volume lifecycle calls.

pkg/hostpath/controllerserver.go
avalluri: Added all these failure checks to the code:
 - NodeSageVolume with no prior ControllerPublishVolume(only if `--enable-attach`)
 - NodePublishVolume with no prior NodeStageVolume
 - NodeUnstageVolume with no prior NodeUnpublishVolume
 - ControllerUnpublishVolume with no prior NodeUn{stage,pulish}Volume
 - DeleteVolume with no prior NodeUn{stage,pulish}Volume and no ControllerUnpublishVolume(only if `--enable-attach`)

pkg/hostpath/controllerserver.go
avalluri: Done.

pkg/hostpath/server.go
```diff
@@ -114,16 +115,64 @@ func parseEndpoint(ep string) (string, string, error) {
...
 	if info.FullMethod == "/csi.v1.Identity/Probe" {
-		return handler(ctx, req)
+		// This call occurs frequently, therefore it only gets log at level 5.
+		pri = 5
+	}
+	glog.V(pri).Infof("GRPC call: %s", info.FullMethod)
+
+	var strippedReq fmt.Stringer
+	v5 := glog.V(5)
+	if v5 {
```
pohly, line 106: Oh, "glog" literally is github.com/golang/glog. Then the check has to be like this.

Another, separate enhancement would be to replace that with klog/v2.

pkg/hostpath/server.go
```diff
@@ -114,16 +115,64 @@ func parseEndpoint(ep string) (string, string, error) {
...
 		glog.Errorf("GRPC error: %v", err)
-	} else {
-		glog.V(5).Infof("GRPC response: %+v", protosanitizer.StripSecrets(resp))
 	}
+
+	if v5 {
+		strippedResp := protosanitizer.StripSecrets(resp)
+		v5.Infof("GRPC response: %+v", strippedResp)
+
+		// convert back to the object, os that logGRPCJson
```
pohly, line 120: Language?

pkg/hostpath/server.go
```diff
@@ -114,16 +115,64 @@ func parseEndpoint(ep string) (string, string, error) {
...
+		json.Unmarshal([]byte(strippedReq.String()), req)
+		json.Unmarshal([]byte(strippedResp.String()), resp)
+		logGRPCJson(info.FullMethod, req, resp, err)
```
pohly, line 123: If this worked, it would modify "resp", the value returned to the caller. We can't do that.

But I also don't think it actually works as intended: `json.Umarshal` just adds values to the object. It does not clear unset ones. That means that secrets remain set.

I think we shouldn't strip secrets in the JSON output. The normal output demonstrates how to do it properly (thus satisfying the goal of showing how to write CSI drivers) and the JSON output is for debugging where we actually may want to see the secrets.

To be on the safe side, we can add a warning to the main README, next to the warning that this driver is just a demo.

pkg/hostpath/controllerserver.go
```diff
@@ -202,6 +202,17 @@ func (hp *hostPath) DeleteVolume(ctx context.Context, req *csi.DeleteVolumeReque
...
 
 	volId := req.GetVolumeId()
+	vol, err := hp.getVolumeByID(volId)
+	if err != nil {
+		// Volume not found: might have already deleted
+		return &csi.DeleteVolumeResponse{}, nil
+	}
+
+	if vol.IsAttached || vol.IsPublished || vol.IsStaged {
+		return nil, status.Errorf(codes.Internal, "Volume '%s' is still used(attached: %v, staged: %v, published: %v) by '%s' node",
```
pohly, line 212: Space before opening bracket, please.

pkg/hostpath/controllerserver.go
```diff
@@ -254,11 +265,91 @@ func (hp *hostPath) ValidateVolumeCapabilities(ctx context.Context, req *csi.Val
...
+	vol, err := hp.getVolumeByID(req.VolumeId)
+	if err != nil {
+		// Not an error: a non-existent volume is not published.
+		// See also https://github.com/kubernetes-csi/external-attacher/pull/165
+		return &csi.ControllerUnpublishVolumeResponse{}, nil
+	}
+
+	// Check to see if the volume is staged/published on a node
+	if vol.IsPublished || vol.IsStaged {
+		return nil, status.Errorf(codes.Internal, "Volume '%s' is still used(staged: %v, published: %v) by '%s' node",
```
pohly, line 343: Space before bracket.

pkg/hostpath/hostpath.go
```diff
@@ -51,6 +51,8 @@ const (
 	storageKind = "kind"
 )
 
+type volumeState string
```
pohly, line 54: Is this used anywhere?

pkg/hostpath/hostpath.go
```diff
@@ -62,23 +64,28 @@ type hostPath struct {
...
+	VolID          string     `json:"volID"`
+	VolSize        int64      `json:"volSize"`
+	VolPath        string     `json:"volPath"`
+	VolAccessType  accessType `json:"volAccessType"`
+	ParentVolID    string     `json:"parentVolID,omitempty"`
+	ParentSnapID   string     `json:"parentSnapID,omitempty"`
+	Ephemeral      bool       `json:"ephemeral"`
+	NodeID         string     `json:"nodeID"`
+	Kind           string     `json:"kind"`
+	ReadOnlyAttach bool       `json:"ReadOnlyAttach"`
```
pohly, line 85: JSON names usually start with lower case.

pohly (changes requested):

pkg/hostpath/server.go
avalluri: > If this worked, it would modify "resp", the value returned to the caller. We can't do that.
> 
> But I also don't think it actually works as intended: `json.Umarshal` just adds values to the object. It does not clear unset ones. That means that secrets remain set.

Yes, you are right. It's my bad I ignored this completely. I should have written the unmarshalled into a new copy.

avalluri: > I think we shouldn't strip secrets in the JSON output. The normal output demonstrates how to do it properly (thus satisfying the goal of showing how to write CSI drivers) and the JSON output is for debugging where we actually may want to see the secrets.
> 
> To be on the safe side, we can add a warning to the main README, next to the warning that this driver is just a demo.

Changed as suggested to log secret fields in JSON format. And also added a README warning.

pkg/hostpath/controllerserver.go
avalluri: Done

pkg/hostpath/controllerserver.go
avalluri: Done.

pkg/hostpath/hostpath.go
avalluri: No, This should be removed. Done

pkg/hostpath/hostpath.go
avalluri: Changed all field names to camelcase.

pohly: I still see `ReadOnlyAttach`?

avalluri: Sorry, I fixed it now.

pohly:
/lgtm

/assign @msau42 

For approval.

cmd/hostpathplugin/main.go
```diff
@@ -41,6 +46,8 @@ var (
 		flag.Var(c, "capacity", "Simulate storage capacity. The parameter is <kind>=<quantity> where <kind> is the value of a 'kind' storage class parameter and <quantity> is the total amount of bytes for that kind. The flag may be used multiple times to configure different kinds.")
 		return c
 	}()
+	enableAttach  = flag.Bool("enable-attach", false, "Enables RPC_PUBLISH_UNPUBLISH_VOLUME capability.")
+	proxyEndpoint = flag.String("proxy-endpoint", "", "Instead of running the CSI driver code, just proxy connections from csiEndpoint to the given listening socket.")
```
msau42, line 50: Do we need some documentation on how to use this proxyEndpoint?

pohly: I can't imagine who besides us in k/k e2e test would want to use this, so I don't think that we need end-user documentation in the README for this. For those who want to know more, there's always the source code...

pkg/hostpath/server.go
pohly: Two reason:
- It's technically complicated because protosanitizer.StripSecrets does not construct new objects, it just wraps the existing ones with a custom String implementation. Therefore a simple `json.Marshal(protosanitizer.StripSecrets(resp))` will still include secrets and more complicated code would be needed.
- This is indeed for verification in mock e2e tests. I don't know whether we have any which look at secrets, but we might, so conceptually it seems better to me to include secrets.

README.md
```diff
@@ -2,6 +2,10 @@
 
 This repository hosts the CSI Hostpath driver and all of its build and dependent configuration files to deploy the driver.
 
+---
+*WARNING: This driver is just a demo implementation.
```
msau42, line 6: I would add that this driver is used for CI testing and has many fake implementations and other non-standard best practices, and should not be used as an example of how to write a real driver.

cmd/hostpathplugin/main.go
msau42: I will disagree that just because something is for internal testing we shouldn't document it (and note it as such). It doesn't have to be elaborate, and even just a pointer to the tests using it could be enough.

pkg/hostpath/server.go
msau42: Thanks, can we document this in the code?

It is also a major change from previous behavior. Would it make sense that we make this a new major release of the hostpath driver with some big warnings that this driver should no longer be used as an example for writing a driver or running in production (however unlikely we think it may be).

pohly: A major version bump indeed seems appropriate. Let's see whether I can add the label as a reminder:

/kind api-change

README.md
pohly: Agreed.

cmd/hostpathplugin/main.go
pohly: Where would you put that?

The main README.md has no documentation whatsoever about driver command line parameters. I'm really not keen on having to document all existing parameters and adding just this one there would be inconsistent. Documenting all parameters is fine and appropriate for the sidecars, but for this driver I still think it is overkill.

README.md
avalluri: I rephrased the warning in the ReadME as @msau42 suggested:

`
*WARNING: This driver is just a demo implementation and is used for CI testing. This has many fake implementations and other non-standard best practices, and should not be used as an example of how to write a real driver.
`

pkg/hostpath/server.go
avalluri: I added this comment to the source code:

> * It's technically complicated because protosanitizer.StripSecrets does not construct new objects, it just wraps the existing ones with a custom String implementation. Therefore a simple `json.Marshal(protosanitizer.StripSecrets(resp))` will still include secrets because it reads fields directly and more complicated code would be needed.
> * This is indeed for verification in mock e2e tests. I don't know whether we have any which look at secrets, but we might, so conceptually it seems better to me to include secrets.

pohly:
Some of the new comments could be improved, but I am okay with letting that stand as it is, so

/lgtm

from me.

cmd/hostpathplugin/main.go
msau42: A comment above this flag is fine with me.

cmd/hostpathplugin/main.go
```diff
@@ -58,7 +65,30 @@ func main() {
 		fmt.Fprintln(os.Stderr, "Deprecation warning: The ephemeral flag is deprecated and should only be used when deploying on Kubernetes 1.15. It will be removed in the future.")
 	}
 
-	driver, err := hostpath.NewHostPathDriver(*driverName, *nodeID, *endpoint, *ephemeral, *maxVolumesPerNode, version, capacity)
+	if *proxyEndpoint != "" {
```
msau42, line 68: General question, it seems like this proxy is standalone and doesn't use any methods of the hostpath driver. Would it make sense to have it be a separate binary?

pohly: We discussed various options when adding this functionality to the mock driver binary; the situation was the same. We considered the whole range (separate repo, separate binary in its own image, separate binary in the same image, new feature in an existing binary and image) and in the end agreed that the last option was the simplest approach.

Everything else comes with additional overhead, both for managing the releases and also regarding image sizes needed during testing (Go runtime only linked onced, same image used for multiple tests).

cmd/hostpathplugin/main.go
pohly: @avalluri can you add this section after "Pre-requisite"?

Features
=======

The driver can provide empty directories that are backed by the same filesystem as EmptyDir volumes. In addition, it can provide raw block volumes that are backed by a single file in that same filesystem and bound to a loop device.

[Various command line parameters](https://github.com/kubernetes-csi/csi-driver-host-path/blob/master/cmd/hostpathplugin/main.go) influence the behavior of the driver. This is relevant in particular for the end-to-end testing that this driver is used for in Kubernetes.

Usually, the driver implements all CSI operations itself. When deployed with the `-proxy-endpoint` parameter, it instead proxies all incoming connections for a CSI driver that is [embedded inside the Kubernetes E2E test suite](https://github.com/kubernetes/kubernetes/tree/master/test/e2e/storage/drivers/csi-test) and used for mocking a CSI driver [with callbacks provided by certain tests](https://github.com/kubernetes/kubernetes/blob/5ad79eae2dcbf33df3b35c48ec993d30fbda46dd/test/e2e/storage/csi_mock_volume.go#L110).

cmd/hostpathplugin/main.go
avalluri: > A comment above this flag is fine with me.

@msau42 Added the comment explaining the usage of this nuew flag.

cmd/hostpathplugin/main.go
avalluri: > @avalluri can you add this section after "Pre-requisite"?
> 
Done. Added the ['Features'  section](https://github.com/kubernetes-csi/csi-driver-host-path/pull/260/commits/8a3b3c522b28ee501277666afff3db0d01f32e2a).

pohly (approved):
/lgtm

msau42:
/approve

Thanks!

PR #262, opened 2021-03-24
deploy: fix CSIStorageCapacity version check

pohly:
**What type of PR is this?**
/kind bug

Checking via "kubectl get" and parsing its output was broken in
combination with "-o pipefail" because then the overall result of the
pipe was false even when the expected error occurred. Checking the
output of "kubectl api-resources" and failing the install when that
command fails (i.e. not using it inside an if) is better.

The check also failed on clusters that had the v1beta1 API but not the
v1alpha1. We need to be very careful about which version we need for
the external-provisioner that is going to be installed. Unfortunately
some heuristics are needed to determine that.

```release-note
Deploying `kubernetes-distributed` did not always correctly detect whether the right CSIStorageCapacity API was supported. Incorrectly enabling the feature then prevented scheduling of pods with late binding volumes.
```

### DIFF
--- a/deploy/kubernetes-distributed/deploy.sh
+++ b/deploy/kubernetes-distributed/deploy.sh
@@ -136,11 +136,21 @@ for component in CSI_PROVISIONER; do
     run kubectl apply -f "${current}"
 done
 
-if kubectl get csistoragecapacities 2>&1 | grep "the server doesn't have a resource type"; then
-    have_csistoragecapacity=false
-else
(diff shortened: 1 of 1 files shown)


pohly:
Both https://github.com/kubernetes-csi/csi-release-tools/pull/139 and changing the Prow job config will fix the test failure on master. I prefer the csi-release-tools update.

In the meantime, let's merge this because it should fix periodic testing on master which already has Bazel usage suppressed.

Issue #251, opened 2021-02-25
Data lost after reboot


daresheep:
Hello,

I had using csi-driver-host-path V1.5 .0

After reboot system,  both pod had been crashed.

describe pods , information this:

```
Events:
  Type     Reason                  Age              From                     Message
  ----     ------                  ----             ----                     -------
  Normal   Scheduled               20s              default-scheduler        Successfully assigned default/virt-launcher-firewall-wlh69 to ceph1
  Normal   SuccessfulAttachVolume  20s              attachdetach-controller  AttachVolume.Attach succeeded for volume "pvc-df463bd3-488b-4e03-b828-5923290f6cdb"
  Warning  FailedMount             0s (x4 over 4s)  kubelet                  MountVolume.SetUp failed for volume "pvc-df463bd3-488b-4e03-b828-5923290f6cdb" : rpc error: code = NotFound desc = volume id d2ec3050-7782-11eb-b03e-46ba88f41811 does not exist in the volumes list

```

After reboot, the mount information were losted, but  discoveryExistingVolumes()  is reading the data form "findmnt", 

This makes all of volume information lost.

Can someone have other idea???

Thank you....

daresheep:
@pohly 

Sir, Can you give me some help, thanks.

pohly:
> I had using csi-driver-host-path V1.5 .0

After the reboot you are still using that version? There were some changes in the code in v1.6.0, but nothing that should have made things worse. Just want to be sure.

Looking at the code, I suspect it was never meant to survive a reboot. Remember, this is a demo driver. It doesn't support all use-cases of a real driver.

Having said that, a PR which enhances the tracking of local volumes and snapshots would be welcome. V1.6.0 introduced capacity simulation, and the size of volumes are known to get lost when restarting the pod.

/help

daresheep:
Thank you for your help!

Just upgrade to V1.6.0  this issue already exist.
 
I think i need to setup other CSI driver to handler this.

Thanks again.

stoneshi-yunify:
I encountered this issue too with latest release (v1.6.2). I looked at the code and I think I've known the reason, the func `discoveryExistingVolumes` can not be used to discover existing volumes after reboot. It can only survive a pod restart, not a node reboot. I managed to get it work by getting the existing volumes from the PersistentVolumes.

[...]

PR #276, opened 2021-04-18
initial implementation of refactor host-path-driver

fengzixu:
**What type of PR is this?**

> /kind bug

For fixing bugs that are mentioned in the issues below

- https://github.com/kubernetes-csi/csi-driver-host-path/issues/251#issuecomment-814719084

Fixes #

```release-note
Fixing bug of host-path-csi-driver which may causes it's failed to start up.
```


pkg/state/snapshot.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+	return false, ""
+}
+
+// getSnapshotPath returns the full path to where the snapshot is stored
+func getSnapshotPath(snapshotID string) string {
+	return filepath.Join(dataRoot, fmt.Sprintf("%s%s", snapshotID, snapshotExt))
+}
+
+func (hps *HostPathDriverState) loadSnapshotsFromFile(snapshotFilePath string) error {
```
pohly, line 92: Instead of `loadSnapshotsFromFile` and `loadVolumesFromFile` and many different files, a single `loadFromFile` and `saveToFile` with *one* state file that contains multiple volumes and snapshots is enough.

`loadFromFile` then only needs to be called once during startup when constructing `HostPathDriverState`. `saveToFile` needs to be called after each change to the in-memory struct.

Don't worry about failure scenarios when writing that file fails, just check and return the error.

This is not a high-quality implementation and should come with a warning that a real driver may have to be more careful about not loosing state, but for this driver it is sufficient.

pkg/state/snapshot.go
```diff
@@ -0,0 +1,117 @@
...
+
+	hps.HostPathSnapshots[snapshotId] = snapshot
+	return snapshot, nil
+}
+
+func (hps *HostPathDriverState) DeleteSnapshot(snapshotId string) error {
+	// Lock before acting on global state. A production-quality
+	// driver might use more fine-grained locking.
+	hps.SnapshotsFileRWLock.Lock()
+	defer hps.SnapshotsFileRWLock.Unlock()
```
pohly, line 68: Locking should be entirely in the gRPC layer, not here.

pkg/state/types.go
```diff
@@ -0,0 +1,110 @@
...
+	storageKind = "kind"
+)
+
+var (
+	// Directory where data for volumes and snapshots are persisted.
+	// This can be ephemeral within the container or persisted if
+	// backed by a Pod volume.
+	dataRoot            = "/csi-data-dir"
+	volumesDataDir      = "/csi-volumes-data"
+	volumesDataFilePath = path.Join(volumesDataDir, "volumes.json")
```
pohly, line 36: These paths must be configurable for unit testing and when running the driver manually as non-root outside of a container.

pkg/state/snapshot.go
fengzixu: > Instead of loadSnapshotsFromFile and loadVolumesFromFile a single loadFromFile and saveToFile 

I agree this point.  Let me improve it. But I don't think aggregate all of logic into the single state.go file is good idea. 

> This is not a high-quality implementation and should come with a warning that a real driver may have to be more careful about not loosing state, but for this driver it is sufficient.

It's true. But I want this refractory change can give a better experience for the new contributor of this project. It is easy for them to understand" What state package did".

>loadFromFile then only needs to be called once during startup when constructing HostPathDriverState. saveToFile needs to be called after each change to the in-memory struct.

Sure

pohly: "The state package manages the internal state of the driver on a node. It ensures that the driver can be stopped and restarted".

It should be that simple. All additional logic for managing snapshots and volumes should be where it resides at the moment.

pkg/state/state.go
```diff
@@ -0,0 +1,421 @@
...
+	snapshot.Path = snapshotFilePath
+	snapshot.CreationTime = creationTime
+	snapshot.SizeBytes = size
+	snapshot.ReadyToUse = true
+
+	hps.HostPathSnapshots[snapshotId] = snapshot
+	return snapshot, nil
+}
+
+func (hps *HostPathDriverState) DeleteSnapshot(snapshotId string) error {
```
fengzixu, line 287: Create/Delete snapshot method doesn't implement "file operation"

pkg/state/state_test.go
```diff
@@ -0,0 +1,256 @@
...
+func TestLoadSnapshotsFromFile(t *testing.T) {
+	err := testHostPathDriverState.loadSnapshotsFromFile()
+	assert.Nil(t, err)
+
+	snapshots, err := testHostPathDriverState.ListSnapshots()
+	assert.Nil(t, err)
+	assert.EqualValues(t, testSnapshots, snapshots)
+}
+
+// func TestLoadStateFromFile(t *testing.T) {
```
fengzixu, line 252: Test should be uncommented after https://github.com/kubernetes-csi/csi-driver-host-path/pull/276/files#r615995654

pkg/state/state.go
```diff
@@ -0,0 +1,421 @@
...
+	}
+
+	return HostPathVolume{}, status.Errorf(codes.NotFound, "volume name %s does not exist in the volumes list", volName)
+}
+
+// CreateVolume allocates capacity, creates the directory for the hostpath volume, and
+// adds the volume to the list.
+//
+// It returns the volume path or err if one occurs. That error is suitable as result of a gRPC call.
+func (hps *HostPathDriverState) CreateVolume(volID, name string, cap int64, volAccessType AccessType, ephemeral bool, kind string, maxVolumeSize int64, capacity Capacity) (hpv *HostPathVolume, finalErr error) {
```
pohly, line 52: The "state" package should not implement the complete logic for creating and manipulating volumes and snapshots. That mixes state management with CSI driver logic.

`HostPathDriverState.AddVolume` should just take a `Volume` struct and then have a few lines of code which update the update the map and store the entire struct in a state file.

PR #277, opened 2021-04-20
dump and restore internal state

pohly:
**What type of PR is this?**
/kind bug

This replaces the previous approach, trying to reconstruct state from
observations, with a simpler dump/restore of the internal state as a
JSON file in the driver's data directory. That old code was broken (did
not distinguish between own volumes and foreign volumes) and
incomplete (not all state restored).

No attempts are made to restore state properly after a node reboot.

https://github.com/kubernetes-csi/csi-driver-host-path/pull/210#discussion_r605734635

```release-note
During startup, the driver may have restored internal state incorrectly (volumes added to internal list that belong to some other driver) or failed to start completely (`failed to get capacity info: no such file or directory`).
```


pkg/hostpath/healthcheck.go
```diff
@@ -48,13 +48,7 @@ type FileSystems struct {
 	Filsystem []ContainerFileSystem `json:"filesystems"`
 }
 
-func getSourcePath(volumeHandle string) string {
-	return fmt.Sprintf("%s/%s", dataRoot, volumeHandle)
-}
```
pohly, line 53: This was a reimplementation of `getVolumePath`. Using `volumeHandle` here instead of `volID` as in the rest of the code confused me ("what is this handle?!"), so I replaced it in the entire file.

pkg/hostpath/healthcheck.go
```diff
@@ -132,14 +126,18 @@ func checkMountPointExist(sourcePath string) (bool, error) {
 	return false, nil
 }
 
-func (hp *hostPath) checkPVCapacityValid(volumeHandle string) (bool, error) {
-	sourcePath := getSourcePath(volumeHandle)
+func (hp *hostPath) checkPVCapacityValid(volID string) (bool, error) {
+	sourcePath := hp.getVolumePath(volID)
```
pohly, line 130: Using `volumeHandle` instead of `volID` as in the rest of the code confused me ("what is this handle?!"), so I replaced it in the entire file.

pkg/hostpath/hostpath.go
```diff
@@ -507,63 +365,12 @@ func loadFromBlockVolume(hostPathVolume hostPathVolume, destPath string) error {
...
-
-func filterVolumeID(sourcePath string) string {
-	volumeSourcePathRegex := regexp.MustCompile(`\[(.*)\]`)
-	volumeSP := string(volumeSourcePathRegex.Find([]byte(sourcePath)))
-	if volumeSP == "" {
-		return ""
-	}
-
-	return strings.TrimSuffix(strings.TrimPrefix(volumeSP, "[/var/lib/csi-hostpath-data/"), "]")
-}
```
pohly, line 549: These two functions are only used by one test, so I moved them there.

pkg/hostpath/controllerserver.go
```diff
@@ -604,33 +601,35 @@ func (hp *hostPath) ListSnapshots(ctx context.Context, req *csi.ListSnapshotsReq
...
-	// case 1: SnapshotId is not empty, return snapshots that match the snapshot id.
+	// case 1: SnapshotId is not empty, return snapshots that match the snapshot id,
+	// none if not found.
 	if len(req.GetSnapshotId()) != 0 {
 		snapshotID := req.SnapshotId
-		if snapshot, ok := hp.snapshots[snapshotID]; ok {
+		if snapshot, err := hp.state.GetSnapshotByID(snapshotID); err == nil {
 			return convertSnapshot(snapshot), nil
 		}
+		return &csi.ListSnapshotsResponse{}, nil
```
pohly, line 611: This `return` here was missing earlier. It only passed because the test is not very precise:
https://github.com/kubernetes-csi/csi-test/blob/a251c44fd49d9eedd55a40b71e5da4ad080ba431/pkg/sanity/controller.go#L1157-L1165

I created an issue for it:
https://github.com/kubernetes-csi/csi-test/issues/335

pkg/hostpath/controllerserver.go
```diff
@@ -604,33 +601,35 @@ func (hp *hostPath) ListSnapshots(ctx context.Context, req *csi.ListSnapshotsReq
...
+	// case 2: SourceVolumeId is not empty, return snapshots that match the source volume id,
+	// none if not found.
 	if len(req.GetSourceVolumeId()) != 0 {
-		for _, snapshot := range hp.snapshots {
+		for _, snapshot := range hp.state.GetSnapshots() {
 			if snapshot.VolID == req.SourceVolumeId {
 				return convertSnapshot(snapshot), nil
 			}
 		}
+		return &csi.ListSnapshotsResponse{}, nil
```
pohly, line 622: Same here.

pohly:
@fengzixu this is what I had in mind: a minimal "state" package whose only responsibility is to store volume and snapshot structs.

The overall patch is quite a bit smaller than what you currently have in https://github.com/kubernetes-csi/csi-driver-host-path/pull/276 despite including some cleanup changes that aren't really necessary (`volumeHandle` -> `volID`).

What do you think? 

/cc @msau42 @xing-yang

fengzixu:
> @fengzixu this is what I had in mind: a minimal "state" package whose only responsibility is to store volume and snapshot structs.
> 
> The overall patch is quite a bit smaller than what you currently have in #276 despite including some cleanup changes that aren't really necessary (`volumeHandle` -> `volID`).
> 
> What do you think?
> 
> /cc @msau42 @xing-yang

Yep. The current implementation is fine for me. It should be enough for fixing the current bug. THanks. Let me close my PR.

pohly:
@xing-yang can you review?

cmd/hostpathplugin/main.go
```diff
@@ -46,6 +46,7 @@ func main() {
 
 	flag.StringVar(&cfg.Endpoint, "endpoint", "unix://tmp/csi.sock", "CSI endpoint")
 	flag.StringVar(&cfg.DriverName, "drivername", "hostpath.csi.k8s.io", "name of the driver")
+	flag.StringVar(&cfg.StateDir, "statedir", "/csi-data-dir", "directory for storing state information across driver restarts, volumes and snapshots")
```
pohly, line 49: Making this configurable is useful because then it is possible again to run the driver as non-root with a state directory in /tmp.

Node operations cannot be tested easily that way (need root for mounting), but even that could be achieved with wrapper scripts that rely on sudo.

pkg/hostpath/healthcheck.go
pohly: Even when both are the same, the cognitive overhead for understanding the source code is higher when some parts use one term and other parts use the other. Everything else uses "volume ID", so this code should do so too.

pkg/hostpath/controllerserver.go
```diff
@@ -604,33 +601,35 @@ func (hp *hostPath) ListSnapshots(ctx context.Context, req *csi.ListSnapshotsReq
...
 	}
 
-	// case 2: SourceVolumeId is not empty, return snapshots that match the source volume id.
+	// case 2: SourceVolumeId is not empty, return snapshots that match the source volume id,
+	// none if not found.
 	if len(req.GetSourceVolumeId()) != 0 {
-		for _, snapshot := range hp.snapshots {
+		for _, snapshot := range hp.state.GetSnapshots() {
 			if snapshot.VolID == req.SourceVolumeId {
 				return convertSnapshot(snapshot), nil
```
xing-yang, line 619: This returns as soon as one snapshot's volume id matches the request, but it is possible that there are multiple snapshots with the same source volume id.  I see that this is in the original code so this can be fixed in a different PR.

pohly: I agree, let's fix this separately. Please file a bug so that we don't forget.

pohly: And also we should have a csi-sanity test for it...

xing-yang:
This looks fine.  I've asked @fengzixu to validate the volume health part.

pkg/hostpath/healthcheck.go
```diff
@@ -190,8 +190,9 @@ func (hp *hostPath) doHealthCheckInControllerSide(volumeHandle string) (bool, st
 	return true, ""
 }
 
-func doHealthCheckInNodeSide(volumeHandle string) (bool, string) {
-	mpExist, err := checkMountPointExist(volumeHandle)
+func (hp *hostPath) doHealthCheckInNodeSide(volID string) (bool, string) {
+	sourcePath := hp.getVolumePath(volID)
```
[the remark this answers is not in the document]
pohly: That change then also should percolate to other places, like `checkMountPointExist`. I've pushed a commit with that change.

Please take another look.

fengzixu:
I tested volume health part based on this PR, test cases as below are fine

1. create volume ---> list volumes ---> get volume health state
2. unmount volume ---> NodeGetVolumeStats/ControllerListVolumeCondition ---> get abnomal volume health state ----> generate event to pod/pvc ```Warning  VolumeConditionAbnormal  4s (x12 over 71m)    csi-pv-monitor-agent-hostpath.csi.k8s.io  The volume isn't mounted```
3. mount volume --->  NodeGetVolumeStats/ControllerListVolumeCondition  ----> get normal volume health state ----> generate event to pod/pvc  ``` Normal    VolumeConditionNormal     pod/my-csi-app                   The Volume returns to the healthy state```

I also tested source path doesn't exist case, it's also fine

Issue #94, opened 2019-09-07
Output container logs when deployment fails


msau42:
Sometimes when testing new features, it causes hostpath driver containers to crash. It would be useful to output the container logs to make debugging easier.

/help

msau42:
/remove-lifecycle stale
@pohly is this fixed?

pohly:
No, we still only have the output of `kubectl describe`: https://github.com/kubernetes-csi/csi-driver-host-path/blob/03f426213c3bcaf0b5d9b16b2ae6b9d08c67112a/deploy/util/deploy-hostpath.sh#L208-L213

PR #281, opened 2021-05-03
deploy fixes

pohly:
**What type of PR is this?**
/kind failing-test

https://k8s-testgrid.appspot.com/sig-storage-csi-ci#distributed-on-master is still failing, the external-provisioner update was incomplete. This showed again that we want the container logs to be visible in CI runs, so this gets added.

Fixes https://github.com/kubernetes-csi/csi-driver-host-path/issues/94

```release-note
NONE
```

### DIFF
--- a/deploy/util/deploy-hostpath.sh
+++ b/deploy/util/deploy-hostpath.sh
@@ -247,10 +247,19 @@ expected_running_pods=6
 cnt=0
 while [ $(kubectl get pods 2>/dev/null | grep '^csi-hostpath.* Running ' | wc -l) -lt ${expected_running_pods} ]; do
     if [ $cnt -gt 30 ]; then
-        echo "$(kubectl get pods 2>/dev/null | grep '^csi-hostpath.* Running ' | wc -l) running pods:"
-        kubectl describe pods
-
-        echo >&2 "ERROR: hostpath deployment not ready after over 5min"
+        echo "Expecting $expected_running_pods, have $(kubectl get pods 2>/dev/null | grep '^csi-hostpath.* Running ' | wc -l)."
+        echo "Deployment:"
+        (set +e; set -x; kubectl describe all,role,clusterrole,rolebinding,clusterrolebinding,serviceaccount,storageclass,csidriver --all-namespaces -l app.kubernetes.io/instance=hostpath.csi.k8s.io)
+        echo
+        echo "Pod logs:"
+        kubectl get pods -l app.kubernetes.io/instance=hostpath.csi.k8s.io --all-namespaces -o=jsonpath='{range .items[*]}{.metadata.name}{" "}{range .spec.containers[*]}{.name}{" "}{end}{"\n"}{end}' | while read -r pod containers; do
+            for c in $containers; do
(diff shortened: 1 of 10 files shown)


deploy/util/deploy-hostpath.sh
```diff
@@ -247,10 +247,19 @@ expected_running_pods=6
...
-        echo "$(kubectl get pods 2>/dev/null | grep '^csi-hostpath.* Running ' | wc -l) running pods:"
-        kubectl describe pods
-
-        echo >&2 "ERROR: hostpath deployment not ready after over 5min"
+        echo "Expecting $expected_running_pods, have $(kubectl get pods 2>/dev/null | grep '^csi-hostpath.* Running ' | wc -l)."
+        echo "Deployment:"
+        (set +e; set -x; kubectl describe all,role,clusterrole,rolebinding,clusterrolebinding,serviceaccount,storageclass,csidriver --all-namespaces -l app.kubernetes.io/instance=hostpath.csi.k8s.io)
+        echo
+        echo "Pod logs:"
+        kubectl get pods -l app.kubernetes.io/instance=hostpath.csi.k8s.io --all-namespaces -o=jsonpath='{range .items[*]}{"\n"}{.metadata.name}{" "}{range .spec.containers[*]}{.name}{" "}{end}{end}' | while read -r pod containers; do
```
jsafrane, line 255: I don't see logs of the driver pod.  `kubectl get pods ...` returns everything:
```
csi-hostpath-attacher-0 csi-attacher 
csi-hostpath-provisioner-0 csi-provisioner 
csi-hostpath-resizer-0 csi-resizer 
csi-hostpath-snapshotter-0 csi-snapshotter 
csi-hostpath-socat-0 socat 
csi-hostpathplugin-0 csi-external-health-monitor-agent csi-external-health-monitor-controller node-driver-registrar hostpath liveness-probe
```

But the loop skips the last line?

jsafrane: The last line returned by `kubectl get pods` does not have `\n` and `read` does not like it?

pohly: Good catch. I checked that logs were printed, but missed the missing ones. Moving `\n` after the output for each pod fixes this.

Even better would be to print "pod name + container name" for each container because then we can avoid one for loop, but I don't know how to do that with jsonpath. If you have a suggestion, please share. In the meantime I'll update the PR with the fix.

jsafrane: I did not know you can have loops in jsonpath until today, so no, I don't have a better one.

pohly:
/retest

Test flake in https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_csi-driver-host-path/281/pull-kubernetes-csi-csi-driver-host-path-1-19-on-kubernetes-1-19/1389506592101109760: "External Storage [Driver: hostpath.csi.k8s.io] [Testpattern: Dynamic PV (filesystem volmode)] multiVolume [Slow] should concurrently access the single volume from pods on the same node"

Issue #192, opened 2020-07-30
Consolidate all the hostpath driver specs into one pod


msau42:
I think we kept attacher separate so that we can easily test with or without it, but I don't see a reason why provisioner, resizer, snapshotter can't all be in the same Pod as the driver. Actually I think our attach required tests are using mock driver, not hostpath driver, so I think it should be safe to bundle attacher in the same pod as well.

pohly:
> I don't see a reason why provisioner, resizer, snapshotter can't all be in the same Pod as the driver

One reason for separate pods is that we can test the RBAC rules for each sidecar separately. For example, if resizer and provisioner both need access to the same resource, but only the RBAC rules for provisioner list that, then resizer will work fine when deployed in the same pod as provisioner, but will fail when deployed separately.

msau42:
Hm that's a good point. I guess it depends on how much we want to balance testing for rbac breakages vs providing a best practices example to the community. I think having all the individual pods:
* is very confusing to someone new to CSI
* isn't something that we recommend in a real production CSI driver
* doesn't follow security best practices since all the controller pieces need to use hostpath to share the socket
* requires each driver to consume a Pod resource per sidecar, which limits the amount of parallel testing we can do in k/k since there's a maximum 110 pods per node, and also could land us in strange scenarios where the driver is only partially able to be deployed on a node.

pohly:
IMHO the main purpose of the csi-hostpath-driver is to facilitate testing. Providing a good example of how to write and deploy a CSI driver is secondary.

> is very confusing to someone new to CSI
> isn't something that we recommend in a real production CSI driver
> doesn't follow security best practices since all the controller pieces need to use hostpath to share the socket

This could be addressed by adding comments and explanations why it is done this way here and what should be done instead. Perhaps link to a "real" CSI driver that does it properly?

> requires each driver to consume a Pod resource per sidecar, which limits the amount of parallel testing we can do in k/k since there's a maximum 110 pods per node, and also could land us in strange scenarios where the driver is only partially able to be deployed on a node.

This is more relevant because it affects testing. But is that really an issue in practice? Have there been test failures because of a partially deployed driver?

pohly:
[...]

PR #282, opened 2021-05-03
single pod deployments

pohly:
**What type of PR is this?**
/kind cleanup

We want to show (and, in Kubernetes, test) deployments with all sidecars and driver in the same pod. We still want to test with individual pods to ensure that RBAC rules work, therefore we need both flavors.

Fixes https://github.com/kubernetes-csi/csi-driver-host-path/issues/192

Based on PR #281 

Depends on an updated prow.sh and new Prow jobs for the non-default deployments.

```release-note
The deploy/kubernetes-x.yy deployments use one pod for all sidecars and the driver. Deploying with separate pods is still supported though the deploy/kubernetes-x.yy-prow deployments, for testing of RBAC rules with Prow.
```


pohly:
/retest

Test flake (?) in https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_csi-driver-host-path/282/pull-kubernetes-csi-csi-driver-host-path-1-21-test-on-kubernetes-1-21/1389899497311571968: "External Storage [Driver: hostpath.csi.k8s.io] [Testpattern: Dynamic PV (default fs)] subPath should support existing directories when readOnly specified in the volumeSource"

PR #288, opened 2021-05-11
WIP: deploy with DaemonSet

pohly:
**What type of PR is this?**
/kind cleanup

DaemonSet has the advantage that "kubectl drain --ignore-daemonsets"
keeps the driver pods running, which is required if normal pods on the
node are to be evicted. When using a StatefulSet, it can happen that
the driver pod gets evicted first and then the other pods cannot be
evicted because their volumes can no longer be unpublished and
unstaged.
    
The downside is that we still have to ensure that the driver only runs
on a single node. With DaemonSet, that is only possible by picking a
node in advance and using that node in a node selector. This node
selection tries to ensure that the pod can really run by checking
various conditions (node ready, network ready, no taints), but this is
less complete than the checks done by kube-scheduler (ignores load).

```release-note
The non-distributed deployments use a DaemonSet to mimick best practices for CSI drivers. The driver still only runs on a single node which has to be chosen in advance based on various criteria (node ready, network ready, no taints) by the deploy script.
```

### DIFF
--- a/deploy/kubernetes-1.18/hostpath/csi-hostpath-plugin.yaml
+++ b/deploy/kubernetes-1.18/hostpath/csi-hostpath-plugin.yaml
@@ -186,7 +186,11 @@ subjects:
 - kind: ServiceAccount
   name: csi-hostpathplugin-sa
 ---
-kind: StatefulSet
+# The driver gets deployed as a DaemonSet because that works better with "kubectl drain":
+# "kubectl drain --ignore-daemonsets" will keep the driver pods running while evicting
+# everything else. This is important because the driver pod is needed while shutting down
+# pods that use its volumes.
+kind: DaemonSet
 apiVersion: apps/v1
 metadata:
   name: csi-hostpathplugin
@@ -196,11 +200,6 @@ metadata:
     app.kubernetes.io/name: csi-hostpathplugin
     app.kubernetes.io/component: plugin
 spec:
-  serviceName: "csi-hostpathplugin"
-  # One replica only:
-  # Host path driver only works when everything runs
-  # on a single node.
-  replicas: 1
   selector:
     matchLabels:
       app.kubernetes.io/instance: hostpath.csi.k8s.io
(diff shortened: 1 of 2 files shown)


pohly:
It's going to be more complex also in other places: the pod name is no longer deterministic, therefore sanity testing will have to be configured differently. The code which updates the test driver config must be updated.

I haven't tackled that yet...

pohly:
/remove-lifecycle stale

@jsafrane do you think using a "proper" DaemonSet is worth the extra complexity or can I close this PR?

Further work would be needed if we want to do this.

jsafrane:
I briefly looked if we can create a Pod that would like a DaemonSet pod without all the shell overhead here. It's probably not possible. kubelet checks Pod owner and even GETs the DaemonSet: 
https://github.com/kubernetes/kubectl/blob/4defba0cec1f594eb410c69bff05b51cddfba8ff/pkg/drain/filters.go#L174-L204

On the other hand, there are other conditions when a pod is skipped during drain:
https://github.com/kubernetes/kubectl/blob/4defba0cec1f594eb410c69bff05b51cddfba8ff/pkg/drain/filters.go#L154-L159

They all seems to be hard to achieve, perhaps our driver Pod could disguise as mirror pod, with a special annotation?
https://github.com/kubernetes/kubectl/blob/4defba0cec1f594eb410c69bff05b51cddfba8ff/pkg/drain/filters.go#L207

jsafrane:
> perhaps our driver Pod could disguise as mirror pod, with a special annotation?

No, that annotation has very specific handling, such a pod is deleted unless it's really a mirror pod.

jsafrane:
Just brainstorming: would it be better to run a "probe" pod, see where it's scheduled and run the driver DaemonSet on the same node?
Alternatively, what is actually goal of the deploy script? All tests run their own driver in go code, without this script. Would it make sense to run the DaemonSet everywhere and use distributed provisioning here?

pohly:
> Just brainstorming: would it be better to run a "probe" pod, see where it's scheduled and run the driver DaemonSet on the same node?

That would be simpler. It's a bit slower (needs to wait for one pod to start), but that shouldn't matter much.

> Alternatively, what is actually goal of the deploy script? All tests run their own driver in go code, without this script. 

The hostpath tests in k/k do that. But all Kubernetes-CSI Prow jobs deploy the hostpath driver once per cluster using deploy.sh, then run the k/k e2e.test with an external driver config.

> Would it make sense to run the DaemonSet everywhere and use distributed provisioning here?

That would massively increase the number of pods during parallel k/k tests. Resource consumption and overloading the cluster are already a problem.

PR #292, opened 2021-05-14
volume lifecycle fixes

### LINKED ISSUE #290: support publishing multiple times and staging exactly once
The driver has traditionally supported publishing multiple times in parallel. That was overlooked when adding a single boolean for `IsPublished`: https://github.com/kubernetes-csi/csi-driver-host-path/blob/7a252595fe60ac7a75fed45214430893aff46d18/pkg/state/state.go#L53

That must be a list of target paths.

pohly:
**What type of PR is this?**
/kind cleanup

The [recently introduced support for tracking of staging and publishing](https://github.com/kubernetes-csi/csi-driver-host-path/pull/260) can be improved:
- Now it properly supports publishing more than once.
- Mutex locking in NodeStageVolume and NodeUnstageVolume was missing.
- NodeUnstageVolume and NodeUnpublishVolume can bail out early
      when nothing needs to be done. In the case of NodeUnpublishVolume
      that avoids unmounting at a path not managed by the driver. The
      driver cannot distinguish between an invalid path and an idempotent
      call, so this is not treated as an error.
- NodePublishVolume checks that the staging target path is valid.
- NodeStageVolume rejects staging more than once.

Fixes #290 

The code that gets enhanced was in no prior release, therefore we don't need to mention this change.
```release-note
NONE
```

### DIFF
--- a/pkg/state/state.go
+++ b/pkg/state/state.go
@@ -48,9 +48,14 @@ type Volume struct {
 	NodeID         string
 	Kind           string
 	ReadOnlyAttach bool
-	IsAttached     bool
-	IsStaged       bool
-	IsPublished    bool
+	Attached       bool
+	// Staged contains the staging target path at which the volume
+	// was staged. A set of paths is used for consistency
+	// with Published.
+	Staged Strings
+	// Published contains the target paths where the volume
+	// was published.
+	Published Strings
 }
 
 type Snapshot struct {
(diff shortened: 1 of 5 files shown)


pohly:
@avalluri : can you perhaps review?

pkg/state/state.go
```diff
@@ -48,9 +48,14 @@ type Volume struct {
 	NodeID         string
 	Kind           string
 	ReadOnlyAttach bool
-	IsAttached     bool
-	IsStaged       bool
-	IsPublished    bool
+	Attached       bool
```
pohly, line 51: I removed the `Is` prefix for consistency with the other booleans and because it no longer works for the fields that aren't a boolean anymore.

avalluri (approved):
Looks good to me.

PR #302, opened 2021-05-21
deploy: fix kustomize issue when using sidecar files for rbac.yaml

pohly:
**What type of PR is this?**
/kind bug

Using kustomize to add labels to a sidecar's local rbac.yaml file
failed because although the file existed already, referencing it via
an absolute path was rejected by kustomize:

```
error: accumulating resources: accumulation err='accumulating resources from '/home/prow/go/src/github.com/kubernetes-csi/external-resizer/deploy/kubernetes/rbac.yaml': security; file '/home/prow/go/src/github.com/kubernetes-csi/external-resizer/deploy/kubernetes/rbac.yaml' is not in or below '/tmp/tmp.ehRnbHpnLX'': new root '/home/prow/go/src/github.com/kubernetes-csi/external-resizer/deploy/kubernetes/rbac.yaml' cannot be absolute
```

```release-note
The deploy scripts did not work when used to test sidecars because kustomize didn't allow rbac.yaml files with absolute paths.
```


pohly:
I have tested the fix locally.

PR #312, opened 2021-06-03
[WIP] Deploy Canary Snapshot Controller When Kubernetes Version in Latest

Kartik494:
**What type of PR is this?**

> /kind bug

Fixes https://github.com/kubernetes-csi/external-snapshotter/issues/516

```release-note

```

### DIFF
--- a/release-tools/prow.sh
+++ b/release-tools/prow.sh
@@ -762,6 +762,8 @@ install_snapshot_crds() {
 
 # Install snapshot controller and associated RBAC, retrying until the pod is running.
 install_snapshot_controller() {
+  TEMP_DIR="$( mktemp -d )"
+  trap 'rm -rf ${TEMP_DIR}' EXIT
   CONTROLLER_DIR="https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/${CSI_SNAPSHOTTER_VERSION}"
   if [[ ${REPO_DIR} == *"external-snapshotter"* ]]; then
       CONTROLLER_DIR="${REPO_DIR}"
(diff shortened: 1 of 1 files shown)


release-tools/prow.sh
```diff
@@ -838,6 +840,31 @@ install_snapshot_controller() {
       echo "kubectl apply -f $SNAPSHOT_CONTROLLER_YAML"
       kubectl apply -f "$SNAPSHOT_CONTROLLER_YAML"
   fi
+  eval current=${SNAPSHOT_CONTROLLER_YAML}
```
pohly, line 843: Why `eval`?

release-tools/prow.sh
```diff
@@ -762,6 +762,8 @@ install_snapshot_crds() {
 
 # Install snapshot controller and associated RBAC, retrying until the pod is running.
 install_snapshot_controller() {
+  TEMP_DIR="$( mktemp -d )"
+  trap 'rm -rf ${TEMP_DIR}' EXIT
   CONTROLLER_DIR="https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/${CSI_SNAPSHOTTER_VERSION}"
   if [[ ${REPO_DIR} == *"external-snapshotter"* ]]; then
```
pohly, line 768: Please update both code paths to use `kubectl --kustomize` and the same code. The only difference between them is which image gets used.

release-tools/prow.sh
```diff
@@ -838,6 +840,31 @@ install_snapshot_controller() {
       echo "kubectl apply -f $SNAPSHOT_CONTROLLER_YAML"
       kubectl apply -f "$SNAPSHOT_CONTROLLER_YAML"
   fi
+  eval current=${SNAPSHOT_CONTROLLER_YAML}
+   if [[ CSI_PROW_KUBERNETES_VERSION == "latest" ]]; then
+      run curl "${current}" --output "${TEMP_DIR}"/snapshot_controller.yaml --silent --location
+    else
+        # Even for local files we need to copy because kustomize only supports files inside
+        # the root of a kustomization.
```
pohly, line 848: How do you know that we have a local file here? The if check doesn't match.

pohly (changes requested):

release-tools/prow.sh
Kartik494: @pohly  Thanks for the suggestions , i will be updating as suggested.

release-tools/prow.sh
Kartik494: @pohly  I used the approach here to store the yaml file of external-snapshotter  in a temp dir and then made a kustomization file that will configure and  patch the image file for master to use canary deployment . The implementation for the same is not yet completed will do the same ASAP . 
 `TEMP_DIR="$( mktemp -d )"` 
` trap 'rm -rf ${TEMP_DIR}' EXIT`

Kartik494:
Closing this PR in reference for this  
https://github.com/kubernetes-csi/external-snapshotter/issues/516

PR #331, opened 2021-09-02
resource exhausted fix

pohly:
**What type of PR is this?**
/kind bug

The driver must return ResourceExhausted when it doesn't have capacity, not
OutOfRange. Then external-provisioner knows that it needs to cause
rescheduling.

This bug was introduced in 6ad318936 via cut-and-paste. For the case above
where the volume is too large for *all* driver instances because it exceeds the
configured maximum volume size it makes sense to return OutOfRange because the
error is permanent.

```release-note
fix rescheduling with distributed provisioning when simulated storage capacity is exhausted
```

### DIFF
--- a/pkg/hostpath/controllerserver.go
+++ b/pkg/hostpath/controllerserver.go
@@ -139,7 +139,7 @@ func (hp *hostPath) CreateVolume(ctx context.Context, req *csi.CreateVolumeReque
 	kind := req.GetParameters()[storageKind]
 	vol, err := hp.createVolume(volumeID, req.GetName(), capacity, requestedAccessType, false /* ephemeral */, kind)
 	if err != nil {
-		return nil, fmt.Errorf("failed to create volume %v: %w", volumeID, err)
+		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


pohly:
/assign @verult 

It would be good to get this into the upcoming release.

pohly:
/hold

This is not quite complete yet: something wraps the ResourceExhausted error, with the result that external-provisioner gets:

```
I0902 17:24:11.614970       1 connection.go:186] GRPC error: rpc error: code = Unknown desc = failed to create volume 9688fa78-0c12-11ec-8595-aad9b8c5520c: rpc error: code = ResourceExhausted desc = requested capacity 53687091200 exceeds remaining capacity for "fast", 100Gi out of 100Gi already used
```

pohly:
/hold cancel

Updated version pushed and tested manually.

I have an E2E test for this particular behavior in PMEM-CSI. Might be worth adding to https://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/csi_mock_volume.go after this fix is in a released version that is used by the Kubernetes CI.

PR #339, opened 2021-11-04
ControllerGetVolume should report abnormal volume condition if volume…

### LINKED ISSUE #338: ControllerGetVolume should report abnormal volume condition if volume is not found
If volume is not found, ControllerGetVolume should return abnormal volume condition rather than return nil.

https://github.com/kubernetes-csi/csi-driver-host-path/blob/v1.7.3/pkg/hostpath/controllerserver.go#L466

zhucan:
… is not found

Signed-off-by: zhucan <zhucan.k8s@gmail.com>

**What type of PR is this?**
/kind bug

ControllerGetVolume should report abnormal volume condition if volume is not found

Fixes #338 

```release-note
ControllerGetVolume should report abnormal volume condition if volume is not found
```

### DIFF
--- a/pkg/hostpath/controllerserver.go
+++ b/pkg/hostpath/controllerserver.go
@@ -463,7 +463,18 @@ func (hp *hostPath) ControllerGetVolume(ctx context.Context, req *csi.Controller
 
 	volume, err := hp.state.GetVolumeByID(req.GetVolumeId())
 	if err != nil {
-		return nil, err
+		// ControllerGetVolume should report abnormal volume condition if volume is not found
+		return &csi.ControllerGetVolumeResponse{
+			Volume: &csi.Volume{
+				VolumeId: req.GetVolumeId(),
+			},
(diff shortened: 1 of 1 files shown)


humblec:
Cc @xing-yang

pkg/hostpath/controllerserver.go
```diff
@@ -463,7 +463,17 @@ func (hp *hostPath) ControllerGetVolume(ctx context.Context, req *csi.Controller
 
 	volume, err := hp.state.GetVolumeByID(req.GetVolumeId())
 	if err != nil {
-		return nil, err
+		return &csi.ControllerGetVolumeResponse{
```
xing-yang, line 466: Can you add a comment here?  Otherwise it is not clear why it is handled this way.

```
# ControllerGetVolume should report abnormal volume condition if volume is not found
```

zhucan: done

PR #342, opened 2022-01-11
Remove unused scripts in hack/ directory

chrishenzie:
**What type of PR is this?**
/kind cleanup

Removes the `hack/` directory because the scripts inside it are unused.

Fixes #

```release-note
NONE
```

@pohly @xing-yang


chrishenzie:
It looks like this URL doesn't work anymore, we'll need to change how we fetch the `csi-sanity` binary

pohly:
> It looks like this URL doesn't work anymore, we'll need to change how we fetch the `csi-sanity` binary

Just build it with `go install`. But where is hack/get-sanity.sh used? Should we perhaps remove it entirely?

The Prow jobs install it through code in prow.sh, which comes from csi-release-tools:
https://github.com/kubernetes-csi/csi-release-tools/blob/4aedf357cac90c044cf3c2b8ff189520e3fd37b2/prow.sh#L231-L241

andyzhangx:
@pohly so this sanity test suite does not work?
I set up sanity test in nfs repo like this, maybe we should switch to this way?
https://github.com/kubernetes-csi/csi-driver-nfs/blob/master/test/sanity/run-test.sh

pohly:
prow.sh works just fine for csi-sanity, no need to change anything there.

What would need further work is `hack/get-sanity.sh` but as it is unused (?) we perhaps should better remove it.

chrishenzie:
Thanks for catching that. I'll update the CSI sanity test suite here: https://github.com/kubernetes-csi/csi-release-tools/pull/182

chrishenzie:
Repurposed this PR to remove all of the `hack/` directory.

`get-sanity.sh` was in use by `e2e-hostpath.sh`, but the e2e testing should be covered by `prow.sh`. The `bump-image-versions.sh` script doesn't appear to be in use either, I couldn't find references to it

mauriciopoppe:
> Edit: I had also removed bump-image-versions.sh but I suspect we may need this

yeah, there might be some scripts that aren't called from this repo but that are called from other repos like test-infra, I saw the same in local volume provisioner so it might be worth looking into the test-infra config jobs too

pohly:
> I had also removed bump-image-versions.sh but I suspect we may need this

Correct. I can be invoked manually to update the YAML files. If that wasn't clear, then adding a section for it under "repository maintenance" or "development" might help to avoid future confusion.

deploy/kubernetes-1.19-test/hostpath/csi-hostpath-snapshotter.yaml
```diff
@@ -37,7 +37,7 @@ spec:
       serviceAccountName: csi-snapshotter
       containers:
         - name: csi-snapshotter
-          image: k8s.gcr.io/sig-storage/csi-snapshotter:v3.0.3
+          image: k8s.gcr.io/sig-storage/csi-snapshotter:v5.0.1
```
pohly, line 40: This might be a change that we shouldn't commit.

@xing-yang: weren't we holding back updating snapshotter on older Kubernetes to get sufficient test coverage for the still supported and recommended 3.x releases?

A comment here in the file would clarify that and serve as a reminder.

xing-yang: I see that this PR tries to update csi-provisioner image to the latest v3.1.0. So similarly it makes sense to also update csi-snapshotter image to the latest v5.0.1 release.

pohly: Where do we then still test csi-snapshotter:v3.x?

chrishenzie: Just updated to leave a comment clarifying this version should not change.

deploy/kubernetes-1.19-test/hostpath/csi-hostpath-snapshotter.yaml
```diff
@@ -37,6 +37,8 @@ spec:
       serviceAccountName: csi-snapshotter
       containers:
         - name: csi-snapshotter
+          # NOTE: This version should not be updated. V3 is still a supported
+          # and recommended version and needs sufficient test coverage.
```
xing-yang, line 41: I see that this is K8s 1.19 when volumesnapshot is still beta. csi-snapshotter v3.0.x supports beta APIs. Snapshot moves to GA in K8s 1.20. So if we are testing K8s 1.19, using v3.0.3 is appropriate.
On the other hand, we discussed about this yesterday and decided V3 support will be dropped so this note here is misleading.

deploy/kubernetes-1.19-test/hostpath/csi-hostpath-provisioner.yaml
```diff
@@ -37,7 +37,7 @@ spec:
       serviceAccountName: csi-provisioner
       containers:
         - name: csi-provisioner
-          image: k8s.gcr.io/sig-storage/csi-provisioner:v2.2.2
+          image: k8s.gcr.io/sig-storage/csi-provisioner:v3.1.0
```
xing-yang, line 40: Why are these images updated to the latest?  v2.2.2 is still supported.  Is testing for v2.2.2 covered elsewhere?
Similar questions for other updated sidecar images.

chrishenzie: It was auto-updated by the script. I can add more notes for these images or alternatively exclude any 1.19 files from being included via `hack/bump-image-versions.sh`

deploy/kubernetes-1.19-test/hostpath/csi-hostpath-snapshotter.yaml
chrishenzie: What version would you prefer this to be? The latest?

deploy/kubernetes-1.19-test/hostpath/csi-hostpath-provisioner.yaml
xing-yang: Our CI tests 3 releases. Currently they are 1.21, 1.22, and 1.23. So I wonder if we should change anything earlier than 1.21.

deploy/kubernetes-1.19-test/hostpath/csi-hostpath-snapshotter.yaml
xing-yang: I'm thinking we should not be updating any images for 1.19 as we are not testing them in our CI any more.

deploy/kubernetes-1.19-test/hostpath/csi-hostpath-provisioner.yaml
pohly: We don't have thorough testing of all supported images.

external-provisioner >= 3.0 is meant to be used with Kubernetes >= 1.20, so this is another image which shouldn't be updated.

chrishenzie:
Updated script to exclude Kubernetes 1.19 and 1.20. Please let me know if this is aligned with your expectations.

deploy/kubernetes-1.19-test/hostpath/csi-hostpath-snapshotter.yaml
xing-yang: @chrishenzie we discussed this at the standup meeting today. We should delete folders for 1.20 and earlier and also update the README after this KEP is merged.  @pohly Please chime in if you want to add anything.

chrishenzie: Ok sounds good. I'll undo the script change. For context, which KEP are you referring to?

xing-yang: Sorry, I meant to say this PR).  We just went through the KEPs targeting 1.24 before I wrote up this comment so I got them mixed up:).

pohly:
> Updated script to exclude paths for Kubernetes 1.19 and 1.20. Please let me know if this is aligned with your expectations.

No :sweat_smile: What we agreed and what @xing-yang wrote above is that the deployment files for any Kubernetes release < 1.21 should be removed and references to them in documentation should be updated.

That reflects that we are no longer maintaining those files on the master branch and cannot guarantee that they work because we stopped testing them.

pohly:
/cancel lgtm

Documentation still needs to be updated.

chrishenzie:
@pohly Can you help me to understand what documentation changes we want to make?

From my understanding, we just need to delete any pre-1.17 markdown docs. Is this correct?

xing-yang:
There are still a few places in the top level README that need to be updated. For example:
https://github.com/kubernetes-csi/csi-driver-host-path/tree/v1.7.3#deployment

chrishenzie:
@xing-yang Updated README to remove references to docs for 1.16 and earlier.

chrishenzie:
@xing-yang @pohly Updated documentation. Please let me know if there are any other doc changes you would like included here.

README.md
```diff
@@ -8,9 +8,9 @@ This repository hosts the CSI Hostpath driver and all of its build and dependent
 
 ## Pre-requisite
 - Kubernetes cluster
-- Running version 1.13 or later
+- Running version 1.17 or later
```
xing-yang, line 11: If we have already deleted folders for v1.20 and lower, should we change this to v1.21 or later?  Then add a line pointing to doc for earlier releases?

xing-yang: Are the examples in the doc from a v1.17 K8s cluster?  If we just change the doc to 1.21 and later, does the example still work?

deploy/kubernetes-1.21-test/hostpath/csi-hostpath-plugin.yaml
```diff
@@ -77,7 +77,7 @@ spec:
           volumeMounts:
           - mountPath: /csi
             name: socket-dir
-          image: k8s.gcr.io/sig-storage/livenessprobe:v2.4.0
+          image: k8s.gcr.io/sig-storage/livenessprobe:v2.5.0
```
xing-yang, line 80: The latest is v2.6.0:
https://github.com/kubernetes-csi/livenessprobe/releases/tag/v2.6.0

deploy/kubernetes-1.21-test/hostpath/csi-hostpath-plugin.yaml
```diff
@@ -97,7 +97,7 @@ spec:
               mountPath: /csi
 
         - name: node-driver-registrar
-          image: k8s.gcr.io/sig-storage/csi-node-driver-registrar:v2.3.0
+          image: k8s.gcr.io/sig-storage/csi-node-driver-registrar:v2.4.0
```
xing-yang, line 100: The latest is 2.5.0:
https://github.com/kubernetes-csi/node-driver-registrar/releases/tag/v2.5.0

deploy/kubernetes-1.21/hostpath/csi-hostpath-plugin.yaml
```diff
@@ -303,13 +303,13 @@ spec:
           volumeMounts:
           - mountPath: /csi
             name: socket-dir
-          image: k8s.gcr.io/sig-storage/livenessprobe:v2.4.0
+          image: k8s.gcr.io/sig-storage/livenessprobe:v2.5.0
```
xing-yang, line 306: 2.6.0

deploy/kubernetes-1.21/hostpath/csi-hostpath-plugin.yaml
```diff
@@ -275,7 +275,7 @@ spec:
               mountPath: /csi
 
         - name: node-driver-registrar
-          image: k8s.gcr.io/sig-storage/csi-node-driver-registrar:v2.3.0
+          image: k8s.gcr.io/sig-storage/csi-node-driver-registrar:v2.4.0
```
xing-yang, line 278: 2.5.0

deploy/kubernetes-distributed/hostpath/csi-hostpath-plugin.yaml
```diff
@@ -60,7 +60,7 @@ spec:
               name: socket-dir
 
         - name: node-driver-registrar
-          image: k8s.gcr.io/sig-storage/csi-node-driver-registrar:v2.3.0
+          image: k8s.gcr.io/sig-storage/csi-node-driver-registrar:v2.4.0
```
xing-yang, line 63: 2.5.0

deploy/kubernetes-distributed/hostpath/csi-hostpath-plugin.yaml
```diff
@@ -132,7 +132,7 @@ spec:
           volumeMounts:
           - mountPath: /csi
             name: socket-dir
-          image: k8s.gcr.io/sig-storage/livenessprobe:v2.4.0
+          image: k8s.gcr.io/sig-storage/livenessprobe:v2.5.0
```
xing-yang, line 135: 2.6.0

README.md
chrishenzie: Ok sounds good

xing-yang:
@chrishenzie We missed updating the snapshot image.
Can you submit a follow-up PR to update the csi-snapshotter image to k8s.gcr.io/sig-storage/csi-snapshotter:v5.0.1?
https://github.com/kubernetes-csi/csi-driver-host-path/blob/master/deploy/kubernetes-1.21/hostpath/csi-hostpath-plugin.yaml#L357

PR #344, opened 2022-01-18
Fix options for volume expansion

gnufied:
Adds new options to control volume expansion.

Also deprecate existing `-node-expand-required` option which was bit weirdly named.


pohly (approved):
/lgtm
/approve

PR #314, opened 2021-06-14
Advertise SINGLE_NODE_MULTI_WRITER capability

chrishenzie:
**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**:
Updates the CSI spec version to v1.5.0, which includes support for `SINGLE_NODE_SINGLE_WRITER` and `SINGLE_NODE_MULTI_WRITER` access modes.

Also advertises the `SINGLE_NODE_MULTI_WRITER` capability so CSI sidecars can supply the appropriate access mode.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:
```release-note
NONE
```

/assign @pohly 
/assign @xing-yang

PR #353, opened 2022-03-28
Add changelog for 1.8.0

jingxu97:
Add changelog

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note

```


xing-yang:
We add the following to the changelog:
https://github.com/kubernetes-csi/csi-driver-host-path/pull/314

pohly (changes requested):
I might have missed some discussions, but don't we first have to update all sidecars, then update the deployment files here, and then release csi-driver-host-path?

At the very least, https://github.com/kubernetes-csi/external-provisioner/pull/710 must get included in the release for Kubernetes 1.24.

jingxu97:
> We should add the following to the changelog: #314

I updated the doc to add this change. Thanks!

pohly:
> I might have missed some discussions, but don't we first have to update all sidecars, then update the deployment files here, and then release csi-driver-host-path?

This PR is another example that shows that just describing the "what" of a change is not enough. I could see from the diff what it was doing, so "Add changelog for 1.8.0" didn't help me understand it.

The relevant missing "why we need this" is that this is the release of csi-driver-host-path for _Kubernetes 1.23_.

PR #356, opened 2022-05-28
fix volume dir mode for unprivileged containers

denisok:
**What type of PR is this?**

/kind bug

Support for unprivileged containers.

Unprivileged container would fail to access pvc. For example:

```yaml
apiVersion: apps/v1
kind: StatefulSet
metadata:
  labels:
    app: perm-test
  name: perm-test
spec:
  replicas: 1
  serviceName: perm-test
  selector:
    matchLabels:
      app: perm-test
  template:
    metadata:
      labels:
        app: perm-test
    spec:
      securityContext:
        fsGroup: 65534
        runAsGroup: 65534
        runAsNonRoot: true
        runAsUser: 65534
      containers:
        - image: busybox
          name: perm-test
          command: ["/bin/sh"]
          args:
            - "-c"
            - |
              touch /mnt/perm_test/file_test && echo passed && sleep 3600 && exit 0
              echo failed
              exit 1
          volumeMounts:
            - mountPath: /mnt/perm_test
              name: perm-test
  volumeClaimTemplates:
    - metadata:
        name: perm-test
      spec:
        accessModes: [ "ReadWriteOnce" ]
        resources:
          requests:
            storage: 1G
```

when applied, it would fail:
```sh
kubectl logs perm-test-0
touch: /mnt/perm_test/file_test: Permission denied
failed
```

Directory that was provisioned has these permissions:
```sh
$ ls -la /mnt/perm_test/
total 8
drwxr-xr-x    2 root     root          4096 May 28 12:42 .
``` 

Unprivileged user (nobody) can't write that. 

https://github.com/kubernetes/minikube/commit/d7bb7c343b61955e30c5830c04d7e83cf40fc37c

similar PR for minikube existed for long time, but the driver there doesn't support multi-node.

https://github.com/kubernetes/minikube/issues/12360

NONE

```release-note
fix volume directory mode for mounted volumes to allow unprivileged container access
```

### DIFF
--- a/pkg/hostpath/hostpath.go
+++ b/pkg/hostpath/hostpath.go
@@ -180,6 +180,13 @@ func (hp *hostPath) createVolume(volID, name string, cap int64, volAccessType st
 		if err != nil {
 			return nil, err
 		}
+		// change dir mode as Mkdir could be affected by umask and Mkdir would inherit acls of the parent dir
+		// default volume type for the root of all volumes is created with DirectoryOrCreate
+		// DirectoryOrCreate is 0755 by default and host root dir for volumes if exist could have different perm (0750)
+		// change volume path mode to 777 to be accessible by unprivileged containers
+		if err = os.Chmod(path, 0777); err != nil {
+			glog.V(4).Infof("Couldn't change volume permissions: %w", err)
+		}
 	case state.BlockAccess:
 		executor := utilexec.New()
 		size := fmt.Sprintf("%dM", cap/mib)


pkg/hostpath/hostpath.go
```diff
@@ -180,6 +180,13 @@ func (hp *hostPath) createVolume(volID, name string, cap int64, volAccessType st
 		if err != nil {
 			return nil, err
 		}
+		// change dir mode as Mkdir could be affected by umask and Mkdir would inherit acls of the parent dir
+		// default volume type for the root of all volumes is created with DirectoryOrCreate
+		// DirectoryOrCreate is 0755 by default and host root dir for volumes if exist could have different perm (0750)
+		// change volume path mode to 777 to be accessible by unprivileged containers
+		if err = os.Chmod(path, 0777); err != nil {
+			glog.V(4).Infof("Couldn't change volume permissions: %w", err)
```
ggriffiths, line 188: from the unit test failure, it looks like you need to use %v here or %s with `err.Error()`:
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_csi-driver-host-path/356/pull-kubernetes-csi-csi-driver-host-path-unit/1532115739618578432#1:build-log.txt%3A126

ggriffiths:
I don't see an issue with this change, but the restriction on privileged containers might be by design. The host path driver is not intended to be used in production or for anything other than testing CSI sidecar functionality.

denisok:
Hi @ggriffiths ,

> I don't see an issue with this change, but the restriction on privileged containers might be by design. 

Maybe just historic? BTW MkDir asks for 0777 anyway, it just wasn't tested that resulting dir actually has that.

Looking to the history of others hostpath drivers that issue was fixed there, minikube is one example, here more:
* https://github.com/kubevirt/hostpath-provisioner/blob/47b2b68b9bbb3b26f5359d90fea6240ac76694be/cmd/provisioner/hostpath-provisioner.go#L250
* https://github.com/rancher/local-path-provisioner/commit/b8bd3b1f412c76e8ed1f38aec5dd9d1e73566627 
* https://github.com/kubernetes/minikube/commit/d7bb7c343b61955e30c5830c04d7e83cf40fc37c

From other point, discussion here:
* https://github.com/kubernetes/kubernetes/pull/39438

But that is for mounting host path to the container and tweaking it's permission. 

For CSI hostpath, it shouldn't be an issue as it is provisioning volumes from the new path that couldn't be mapped to the host path.

Better way is probably to implement fsGroup and other SecurityContext features.

So far as those volumes paths are managed by csi driver it looks like this change is safe.

> The host path driver is not intended to be used in production or for anything other than testing CSI sidecar functionality.

Sure, I understand about production, as about "anything other than testing CSI sidecar" maybe extending to other tools that use csi hostpath in testing - would extend testing for CSI sidecar? :)

Thanks,
Denis

denisok:
Hi @ggriffiths @humblec , please review :)

humblec:
I doubt giving worldwide permission (777) is the right fix. imo, we have to make use of `fsgroup` in the pod spec and let kubelet take care of the volume permissions based on the `fsgroupchangepolicy` ( https://github.com/kubernetes-csi/csi-driver-host-path/blob/master/deploy/kubernetes-1.21/hostpath/csi-hostpath-driverinfo.yaml). 

Cc @gnufied @jsafrane @pohly

jsafrane:
Yes, users should use fsGroup to access their volumes. Most of CSI drivers do just mkfs.ext4, which leaves `/` accessible only to root.

denisok:
@jsafrane I agree that fsGroup is the right way to go, but currently that feature is absent in this csi driver.

Also you are talking about `mkfs` so you probably talking about block allocation, this change is only about directory mount and creating that with different permissions.

jsafrane:
> @jsafrane I agree that fsGroup is the right way to go, but currently that feature is absent in this csi driver.

The driver does not need to implement anything, IMO. Just the CSIDriver instance should not rely on our broken but backward compatible heuristic and declare `fsGroupPolicy: File`.

I tested this CSIDriver and your pod can read/write to /mnt/perm_test:

```
apiVersion: storage.k8s.io/v1
kind: CSIDriver
metadata:
    name: hostpath.csi.k8s.io
spec:
    attachRequired: true
    fsGroupPolicy: File
    podInfoOnMount: true
    requiresRepublish: false
    storageCapacity: false
    volumeLifecycleModes:
    - Persistent
    - Ephemeral
```

```
$  kubectl exec -ti perm-test-0 sh
/ $ id
uid=65534(nobody) gid=65534(nobody) groups=65534(nobody)
/ $ cd /mnt/perm_test/
/mnt/perm_test $ touch aaa
/mnt/perm_test $ ls -la
total 0
drwxrwsr-x    1 root     nobody          24 Jun 22 09:14 .
drwxr-xr-x    1 root     root            18 Jun 22 09:14 ..
-rw-r--r--    1 nobody   nobody           0 Jun 22 09:14 aaa
-rw-r--r--    1 nobody   nobody           0 Jun 22 09:14 file_test
```

denisok:
yep, looks like this patch is definitely incorrect. It works as kubelet and csi driver do nothing passing the volume to container, but in reality CSIDriver or kubelet should set right permissions.

https://github.com/kubernetes-csi/docs/blob/master/book/src/support-fsgroup.md

@jsafrane could you please help me to further understand how it works and how to correctly configure it.

So `fsGroupPolicy` is not set by CSIDriver spec and thus it is defaults to `ReadWriteOnceWithFSType`, `accessModes` contains `ReadWriteOnce` in example above, but the `fsType` is not set and that is probably why permissions don't get propagated.

Should default spec be set to `fsGroupPolicy: File` and offload permissions set to Kubernetes? Or should CSIDriver support `fsType` some how?

Thanks for helping to understand how it works/should work!

jsafrane:
> Should default spec be set to fsGroupPolicy: File and offload permissions set to Kubernetes? Or should CSIDriver support fsType some how?

the best would be to have `fsGroupPolicy: File` in CSIDriver instance, that's exactly why the field is there.

PR #374, opened 2022-11-02
v1 snapshot object has to be used instead of alpha and beta in deployment

humblec:
- [x] alpha snapshot object deleted
- [x] v1 snapshot object has been added and updated the doc to use the same
  
> /kind cleanup

```release-note
NONE
```


pohly (approved):
/lgtm

PR #381, opened 2022-11-24
Enforce SINGLE_NODE_SINGLE_WRITER access mode in NodePublishVolume

### LINKED ISSUE #380: Implement support for SINGLE_NODE_SINGLE_WRITER enforcement in NodePublishVolume
See the second table in this section for more details:
https://github.com/container-storage-interface/spec/blob/v1.7.0/spec.md#nodepublishvolume

This is a requirement for beta graduation of the `ReadWriteOncePod` feature:
https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/2485-read-write-once-pod-pv-access-mode#beta

chrishenzie:
**What type of PR is this?**
/kind feature

This is a first draft of adding support for `SINGLE_NODE_SINGLE_WRITER` access mode enforcement in the hostpath driver.

See the second table in this section for more details:
https://github.com/container-storage-interface/spec/blob/v1.7.0/spec.md#nodepublishvolume

**NOTE:** This is hacky. Any suggestions for improvement are much appreciated.

Fixes #380 

```release-note
Enforces volumes the with SINGLE_NODE_SINGLE_WRITER access mode can only be mounted at one target path at a time
```

**Testing**:
This was tested using the k8s `e2e.test` binary to install the (custom) hostpath driver and run pods on the same host.
```sh
make WHAT=test/e2e/e2e.test

_output/local/bin/linux/amd64/e2e.test \
  --kubeconfig="${HOME}/.kube/config" \
  -ginkgo.focus='csi-hostpath.*should block a second pod from using an in-use ReadWriteOncePod volume on the same node'
```

/sig storage
/cc @msau42
/cc @jsafrane

### DIFF
--- a/pkg/hostpath/nodeserver.go
+++ b/pkg/hostpath/nodeserver.go
@@ -31,10 +31,13 @@ import (
 	"k8s.io/utils/mount"
 )
 
-const TopologyKeyNode = "topology.hostpath.csi/node"
+const (
+	TopologyKeyNode = "topology.hostpath.csi/node"
+
+	failedPreconditionAccessModeConflict = "volume uses SINGLE_NODE_SINGLE_WRITER access mode and is already mounted at a different target path"
+)
 
 func (hp *hostPath) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolumeRequest) (*csi.NodePublishVolumeResponse, error) {
-
 	// Check arguments
 	if req.GetVolumeCapability() == nil {
 		return nil, status.Error(codes.InvalidArgument, "Volume capability missing in request")
@@ -80,6 +83,10 @@ func (hp *hostPath) NodePublishVolume(ctx context.Context, req *csi.NodePublishV
 	vol, err := hp.state.GetVolumeByID(req.GetVolumeId())
 	if err != nil {
 		return nil, status.Error(codes.NotFound, err.Error())
+	}
+
+	if hasSingleNodeSingleWriterAccessMode(req) && isMountedElsewhere(req, vol) {
+		return nil, status.Error(codes.FailedPrecondition, failedPreconditionAccessModeConflict)
 	}
 
 	if !ephemeralVolume {
@@ -520,4 +527,22 @@ func makeFile(pathname string) error {
 		}
 	}
 	return nil
+}
+
+// hasSingleNodeSingleWriterAccessMode checks if the publish request uses the
+// SINGLE_NODE_SINGLE_WRITER access mode.
+func hasSingleNodeSingleWriterAccessMode(req *csi.NodePublishVolumeRequest) bool {
+	accessMode := req.GetVolumeCapability().GetAccessMode()
+	return accessMode != nil && accessMode.GetMode() == csi.VolumeCapability_AccessMode_SINGLE_NODE_SINGLE_WRITER
+}
+
+// isMountedElsewhere checks if the volume to publish is mounted elsewhere on
+// the node.
+func isMountedElsewhere(req *csi.NodePublishVolumeRequest, vol state.Volume) bool {
+	for _, targetPath := range vol.Published {
+		if targetPath != req.GetTargetPath() {
+			return true
+		}
+	}
+	return false
 }
--- a/pkg/hostpath/nodeserver.go
+++ b/pkg/hostpath/nodeserver.go
@@ -60,6 +60,8 @@ func (hp *hostPath) NodePublishVolume(ctx context.Context, req *csi.NodePublishV
 	hp.mutex.Lock()
 	defer hp.mutex.Unlock()
 
+	mounter := mount.New("")
+
 	// if ephemeral is specified, create volume here to avoid errors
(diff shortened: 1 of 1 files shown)


pkg/hostpath/nodeserver.go
```diff
@@ -522,3 +538,34 @@ func makeFile(pathname string) error {
...
+
+// hasSingleNodeSingleWriterAccessMode checks if the publish request uses the
+// SINGLE_NODE_SINGLE_WRITER access mode.
+func hasSingleNodeSingleWriterAccessMode(req *csi.NodePublishVolumeRequest) bool {
+	accessMode := req.GetVolumeCapability().GetAccessMode()
+	return accessMode != nil && accessMode.GetMode() == csi.VolumeCapability_AccessMode_SINGLE_NODE_SINGLE_WRITER
+}
+
+// TODO: WIP: Fix this hacky approach. Does this work on windows?
+func parsePVCName(req *csi.NodePublishVolumeRequest) (string, error) {
```
jsafrane, line 550: I think it's PV name what is in the mount path, not PVC name. And yes, it's very hacky.

jsafrane: The driver may already maintain list of mounts, at least there is `vol.Published.Add(targetPath)` - maybe you can check from there. The list will be empty after driver restart, still, this is not a production CSI driver (I hope).
`

chrishenzie: The target paths in `vol.Published` contain pod IDs -- this solution would require performing substring matches against the entries in `vol.Published` (or I could have a separate cache of PV names). I was thinking the mounter approach might resemble a more "production" use-case that could be mimicked in other CSI drivers, and has the advantage of being stateless, but is likely less performant.

A drawback is this solution is tightly coupled to K8s and Linux. What are your thoughts on this? I could make this PV parsing logic a little safer by only enforcing the access mode if the target path matches some regex pattern, etc. Any suggestions are appreciated.

jsafrane: A real CSI driver will do `mounter.GetMountRefs`, because there is a device or a NFS share. I am not sure it makes sense to mimic anything real in the csi-driver-hostpath.

chrishenzie: Updated to scan entries in `vol.Published`

pkg/hostpath/nodeserver.go
```diff
@@ -522,3 +534,30 @@ func makeFile(pathname string) error {
...
+
+// isMountedElsewhere checks if the volume to publish is mounted elsewhere on
+// the node.
+func isMountedElsewhere(req *csi.NodePublishVolumeRequest, pvName string, volume state.Volume) bool {
+	for _, pub := range volume.Published {
+		if strings.Contains(pub, pvName) && pub != req.GetTargetPath() {
+			return true
+		}
+	}
+	return false
```
jsafrane, line 562: I meant `volume.Published` is list of NodePublish `targetPaths`: https://github.com/kubernetes-csi/csi-driver-host-path/blob/4191fcd0bd787154f0fe1d6afcf78dcf44796d4f/pkg/hostpath/nodeserver.go#L192

You can remove `parsePVName` and filter by `targetPath` only:
```go
	for _, pub := range vol.Published {
		if pub != req.GetTargetPath() {
			return true
		}
	}
```

chrishenzie: Aha I see, because we're using `SINGLE_NODE_SINGLE_WRITER` there should only be one target path in `vol.Published`. Thanks!

chrishenzie:
/retest

Failed due to:
```
{"msg":"PASSED External Storage [Driver: hostpath.csi.k8s.io] [Testpattern: Generic Ephemeral-volume (default fs) (immediate-binding)] ephemeral should create read-only inline ephemeral volume","total":-1,"completed":1,"skipped":279,"failed":1,"failures":["External Storage [Driver: hostpath.csi.k8s.io] [Testpattern: Dynamic PV (default fs)] capacity provides storage capacity information"]}
{"msg":"FAILED External Storage [Driver: hostpath.csi.k8s.io] [Testpattern: Dynamic PV (default fs)] capacity provides storage capacity information","total":-1,"completed":0,"skipped":183,"failed":1,"failures":["External Storage [Driver: hostpath.csi.k8s.io] [Testpattern: Dynamic PV (default fs)] capacity provides storage capacity information"]}
```
```
[It] provides storage capacity information
  test/e2e/storage/testsuites/capacity.go:112
[1m[0m: Creating a StorageClass
Dec  7 02:13:55.775: FAIL: Timed out after 60.006s.
after creating storage class
no CSIStorageCapacity objects for storage class "capacity-5029-e2e-sc6xqbw"
```

chrishenzie:
Bumping this, is there any other feedback I can address?

jsafrane:
/lgtm
/approve
sorry about the delay, too much work to do

chrishenzie:
Np, thx for the review!

PR #395, opened 2023-02-16
fix: set fstype when access type is mount

MeinhardZhou:
**What type of PR is this?**
/kind bug

when volume access is `mount`, we should mount with fstype, but here didn't set.

```release-note
none
```

### DIFF
--- a/pkg/hostpath/nodeserver.go
+++ b/pkg/hostpath/nodeserver.go
@@ -182,7 +182,7 @@ func (hp *hostPath) NodePublishVolume(ctx context.Context, req *csi.NodePublishV
 		}
 		path := hp.getVolumePath(volumeId)
 
-		if err := mounter.Mount(path, targetPath, "", options); err != nil {
+		if err := mounter.Mount(path, targetPath, fsType, options); err != nil {
 			var errList strings.Builder
 			errList.WriteString(err.Error())
 			if vol.Ephemeral {


pkg/hostpath/nodeserver.go
```diff
@@ -182,7 +182,7 @@ func (hp *hostPath) NodePublishVolume(ctx context.Context, req *csi.NodePublishV
 		}
 		path := hp.getVolumePath(volumeId)
 
-		if err := mounter.Mount(path, targetPath, "", options); err != nil {
+		if err := mounter.Mount(path, targetPath, fsType, options); err != nil {
```
pohly, line 185: Why has this not caused tests to fail?

Do we need more or better tests?

MeinhardZhou: @pohly

nixpanic: I don't think this makes a difference. The command executes `mount -obind ...`, which creates a new mountpoint from an existing directory. The filesystem of the final mountpoint (`targetPath`) will always be that of the original directory (`path`).

MeinhardZhou: > I don't think this makes a difference. The command executes `mount -obind ...`, which creates a new mountpoint from an existing directory. The filesystem of the final mountpoint (`targetPath`) will always be that of the original directory (`path`).

Yes, I found out that `targetPath` depends on the filesystem type of the original volume `path`. `mount --bind` is like a hard link. Maybe I'll consider other code to fix `fstype` not working.

nixpanic: You can not really change the fstype. The `path` is located on the host, and already has a fstype even before the driver is deployed. If  a special fstype is requested, you could return an error if it does not match the filesystem on `path`, but that may cause breakage where the driver is used now (test suites only?).

Or, a more complex solution would be to create a loopback device, and format that with the requested fstype. I am not sure it is worth the effort to go that route though.

PR #399, opened 2023-03-10
Add GroupControllerServer interface for VolumeGroupSnapshot API

nixpanic:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature

> /kind flake

VolumeGroupSnapshots have been added to the CSI Spec with version 1.8.0. This PR implements the new API.

Fixes #

The Kubernetes CSI external-snapshotter does not have the full functionality at this time. The implementation of the VolumeGroupSnapshot functions is largely untested at the moment.

```release-note
Together with the Kubernetes CSI external-snapshotter, the csi-driver-host-path provides ALPHA support for the VolumeGroupSnapshot functionality.
```


pkg/hostpath/controllerserver.go
```diff
@@ -763,6 +756,236 @@ func (hp *hostPath) ControllerExpandVolume(ctx context.Context, req *csi.Control
...
+	}
+	// Check arguments
+	if len(req.GetSourceVolumeIds()) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "SourceVolumeIds missing in request")
+	}
+
+	// Lock before acting on global state. A production-quality
+	// driver might use more fine-grained locking.
+	hp.mutex.Lock()
+	defer hp.mutex.Unlock()
```
xing-yang, line 788: I see that mutex is used for every operation, i.e., create snapshot, delete snapshot.  This could cause race condition?

pkg/hostpath/controllerserver.go
```diff
@@ -763,6 +756,236 @@ func (hp *hostPath) ControllerExpandVolume(ctx context.Context, req *csi.Control
...
+	if len(req.GetGroupSnapshotId()) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "GroupSnapshot ID missing in request")
+	}
+
+	groupSnapshotID := req.GetGroupSnapshotId()
+
+	// Lock before acting on global state. A production-quality
+	// driver might use more fine-grained locking.
+	hp.mutex.Lock()
+	defer hp.mutex.Unlock()
```
xing-yang, line 910: Similar question about the mutex.

pkg/state/state.go
```diff
@@ -113,11 +122,31 @@ type State interface {
 	// snapshot ID. It is not an error when such a snapshot
 	// does not exist.
 	DeleteSnapshot(snapshotID string) error
+
+	// GetGroupSnapshotByID retrieves a snapshot by its unique ID or
```
xing-yang, line 126: snapshot -> group snapshot

pkg/state/state.go
```diff
@@ -113,11 +122,31 @@ type State interface {
...
+	// returns an error including that ID when not found.
+	GetGroupSnapshotByID(vgsID string) (GroupSnapshot, error)
+
+	// GetGroupSnapshotByName retrieves a snapshot by its name or
```
xing-yang, line 130: snapshot -> group snapshot

pkg/state/state.go
```diff
@@ -113,11 +122,31 @@ type State interface {
...
+	// returns an error including that name when not found.
+	GetGroupSnapshotByName(volName string) (GroupSnapshot, error)
+
+	// GetGroupSnapshots returns all currently existing snapshots.
```
xing-yang, line 134: snapshots -> group snapshots

pkg/state/state.go
```diff
@@ -113,11 +122,31 @@ type State interface {
...
+	GetGroupSnapshots() []GroupSnapshot
+
+	// UpdateGroupSnapshot updates the existing hostpath snapshot,
```
xing-yang, line 137: snapshot -> group snapshot

pkg/state/state.go
```diff
@@ -113,11 +122,31 @@ type State interface {
...
+	// identified by its snapshot ID, or adds it if it does not exist yet.
```
xing-yang, line 138: snapshot -> group snapshot

pkg/state/state.go
```diff
@@ -113,11 +122,31 @@ type State interface {
...
+	UpdateGroupSnapshot(snapshot GroupSnapshot) error
+
+	// DeleteGroupSnapshot deletes the snapshot with the given
```
xing-yang, line 141: snapshot -> group snapshot

pkg/state/state.go
```diff
@@ -113,11 +122,31 @@ type State interface {
...
+	// snapshot ID. It is not an error when such a snapshot does not exist.
```
xing-yang, line 142: snapshot -> group snapshot

pkg/hostpath/controllerserver.go
nixpanic: The mutex is already used for all other CSI RPC calls, they only call it on the top-level function. There are no helper functions that use the mutex to lock the `hostPath.state`.

xing-yang: Ok

nixpanic:
In case someone wants to test this PR, a container image is available at `quay.io/nixpanic/csi-driver-host-path:GroupControllerServer`.

xing-yang:
This missed adding a `Type: csi.PluginCapability_Service_GROUP_CONTROLLER_SERVICE` here:
https://github.com/kubernetes-csi/csi-driver-host-path/blob/v1.11.0/pkg/hostpath/identityserver.go#L54

RaunakShah (commented):
Also, `GetPluginCapabilities` in `hostpath/identityserver.go` needs to return the group controller capability

nixpanic:
Thanks for the reviews @xing-yang and @RaunakShah! Your comments should have been addressed now.

The PR has been rebased so that conflicts with `go.mod` from the master branch are resolved too.

nixpanic:
A container-image with the current PR is at `quay.io/nixpanic/csi-driver-host-path:GroupControllerServer_v2`

nixpanic:
FYI, tests will fail until kubernetes-csi/csi-test#450 is merged and used by the CI.

pkg/hostpath/controllerserver.go
```diff
@@ -763,6 +756,236 @@ func (hp *hostPath) ControllerExpandVolume(ctx context.Context, req *csi.Control
 	}, nil
 }
 
+func (hp *hostPath) GroupControllerGetCapabilities(context.Context, *csi.GroupControllerGetCapabilitiesRequest) (*csi.GroupControllerGetCapabilitiesResponse, error) {
```
xing-yang, line 759: Can you move these GroupController functions into a new file called `groupcontrollerserver.go`?  This will make it a little easier to review them in the future.

nixpanic: Sure! That has been done now.

pkg/hostpath/controllerserver.go
```diff
@@ -601,6 +589,11 @@ func (hp *hostPath) DeleteSnapshot(ctx context.Context, req *csi.DeleteSnapshotR
 	hp.mutex.Lock()
 	defer hp.mutex.Unlock()
 
+	// If the snapshot has a GroupSnapshotID, deletion is not allowed and should return InvalidArgument.
+	if snapshot, err := hp.state.GetSnapshotByID(snapshotID); err != nil && snapshot.GroupSnapshotID != "" {
+		return nil, status.Errorf(codes.InvalidArgument, "Snapshot with ID %s is part of snapshotgroup %s", snapshotID, snapshot.GroupSnapshotID)
```
RaunakShah, line 594: part of snapshotgroup -> part of groupsnapshot

pkg/hostpath/groupcontrollerserver.go
```diff
@@ -0,0 +1,273 @@
...
+
+	// Lock before acting on global state. A production-quality
+	// driver might use more fine-grained locking.
+	hp.mutex.Lock()
+	defer hp.mutex.Unlock()
+
+	// Need to check for already existing groupsnapshot name, and if found check for the
+	// requested sourceVolumeIds and sourceVolumeIds of groupsnapshot that has been created.
+	if exGS, err := hp.state.GetGroupSnapshotByName(req.GetName()); err == nil {
+		// Since err is nil, it means the groupsnapshot with the same name already exists need
```
RaunakShah, line 67: nit: `already exists need` -> `already exists. Need ...`

pkg/hostpath/groupcontrollerserver.go
```diff
@@ -0,0 +1,273 @@
...
+		// to check if the sourceVolumeIds of existing groupsnapshot is the same as in new request.
+
+		if !exGS.MatchesSourceVolumeIDs(req.GetSourceVolumeIds()) {
+			return nil, status.Errorf(codes.AlreadyExists, "snapshot with the same name: %s but with different SourceVolumeIds already exist", req.GetName())
```
RaunakShah, line 71: snapshot -> group snapshot

pkg/state/state.go
RaunakShah: Maybe I'm looking at an older commit, but I think this still needs to be addressed.

pkg/state/state.go
```diff
@@ -262,3 +292,73 @@ func (s *state) DeleteSnapshot(snapshotID string) error {
...
+
+func (gs *GroupSnapshot) MatchesSourceVolumeIDs(sourceVolumeIDs []string) bool {
+	if len(gs.SnapshotIDs) != len(sourceVolumeIDs) {
+		return false
+	}
+
+	for _, snapshotID := range gs.SnapshotIDs {
+		found := false
+
+		for _, sourceVolumeID := range sourceVolumeIDs {
```
RaunakShah, line 351: How about, in a single loop creating a map of existing `snapshotID`s and then verifying if `sourceVolumeIDs` exists in that map?

pkg/state/state_test.go
```diff
@@ -124,3 +124,34 @@ func TestSnapshotsFromSameSource(t *testing.T) {
...
+func TestVolumeGroupSnapshots(t *testing.T) {
+	tmp := t.TempDir()
+	statefileName := path.Join(tmp, "state.json")
+
+	s, err := New(statefileName)
+	require.NoError(t, err, "construct state")
+	require.Empty(t, s.GetGroupSnapshots(), "initial groupsnapshots")
+
+	_, err = s.GetGroupSnapshotByID("foo")
+	require.Equal(t, codes.NotFound, status.Convert(err).Code(), "GetSnapshotByID of non-existent groupsnapshot")
```
RaunakShah, line 137: GetSnapshotByID -> GetGroupSnapshotByID

nixpanic:
Thanks @RaunakShah, your comments should be addressed now.

ggriffiths (changes requested):
Looks like the builds are failing here, feel free to re-request review when this is ready.

leonardoce:
Hi. I'm testing this PR (thanks, @xing-yang, for pointing me at this).
I realized that Group Snapshots are failing with `failed to take group snapshot of the volumes [...]: “rpc error: code = AlreadyExists desc = group snapshot with the same name: ... but with different SourceVolumeIds already exist`).

In the code, it looks like we're comparing PV handle IDs against snapshot handle IDs (see the [`MatchesSourceVolumeIDs` function](https://github.com/kubernetes-csi/csi-driver-host-path/pull/399/files#diff-7e4e87a1e777423a3d9a915e00fb78a9c4fc2f4c476863f60b67e89b9c635697R347) and the [`CreateVolumeGroupSnapshot` function](https://github.com/kubernetes-csi/csi-driver-host-path/pull/399/files#diff-2b4d910b5749c660750b11299fd3372c5b4a0b61e9cb9e4c363df7b1dcb98682R142)).

Doing this, the first `CreateVolumeGroupSnapshot` call will work correctly since the internal state is empty, and the subsequent calls will fail. I had positive results adding [this commit](https://github.com/leonardoce/csi-driver-host-path/commit/78e0b3dccf30faaccafed95cd61217b38f0d7c24) that makes the `GroupSnapshot` struct keep track of both snapshot IDs and source volumes IDs.

Happy to help. Thank you!

leonardoce:
I'd also consider [this commit](https://github.com/leonardoce/csi-driver-host-path/commit/d7e25e1a247b18ef261721da054e96fe653ad3f8), setting the `GroupSnapshotId` field in the `ListSnapshots` GRPC response

nixpanic:
> I'd also consider [this commit](https://github.com/leonardoce/csi-driver-host-path/commit/d7e25e1a247b18ef261721da054e96fe653ad3f8), setting the `GroupSnapshotId` field in the `ListSnapshots` GRPC response

Thanks! Will include that in the next update for this PR.

nixpanic:
Thanks @leonardoce, your commits have been added in test-only PR #426 now.

nixpanic:
Ah! tests fail as `release-tools/prow.sh` still uses `CSI_PROW_SANITY_VERSION v5.0.0`. That needs to be v5.1.0 or newer.

This driver reports the new capability, bit the v5.0.0 version of the tests do not know about it.
image

xing-yang:
Submitted this PR to update release-tools: https://github.com/kubernetes-csi/csi-driver-host-path/pull/474

PR #405, opened 2023-04-12
Update dependency go modules for k8s v1.27.0

sunnylovestiramisu:
Ran kubernetes-csi/csi-release-tools go-get-kubernetes.sh -p 1.27.0


go.mod
```diff
@@ -41,60 +41,60 @@ require (
...
 )
 
-replace k8s.io/api => k8s.io/api v0.27.0-rc.0
+replace k8s.io/api => k8s.io/api v0.27.0
 
-replace k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.27.0-rc.0
+replace k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.27.0
 
-replace k8s.io/apimachinery => k8s.io/apimachinery v0.27.0-rc.0
+replace k8s.io/apimachinery => k8s.io/apimachinery v0.28.0-alpha.0
```
xing-yang, line 52: This one goes to 0.28?

xing-yang: Does the script always go to the latest release, even if it is alpha?

sunnylovestiramisu: Seems like it, it is the script we have always been using.

PR #430, opened 2023-06-26
chore: use copy(to, from) instead of a loop

testwill:
**What type of PR is this?**
> /kind cleanup

should use copy(to, from) instead of a loop

Fixes #

```release-note
NONE
```


pohly (commented):
/ok-to-test

PR #456, opened 2023-10-06
Update readme doc

astraw99:
**What type of PR is this?**
/kind documentation

Fixed and updated the readme doc, to make it better to test locally.
Tested with the latest code, safe to merge.

```release-note
NONE
```

### DIFF
--- a/pkg/hostpath/README.md
+++ b/pkg/hostpath/README.md
@@ -7,9 +7,15 @@
 $ make
 ```
 
+### Set endpoint
+```
+$ endpoint=unix:///tmp/csi.sock # unix (default)
+$ #endpoint=tcp://127.0.0.1:10000 # tcp
+```
+
 ### Start Hostpath driver
 ```
-$ sudo ./bin/hostpathplugin --endpoint tcp://127.0.0.1:10000 --nodeid CSINode -v=5
+$ sudo ./bin/hostpathplugin --endpoint $endpoint --nodeid CSINode -v=5
 ```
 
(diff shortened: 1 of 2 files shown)


pkg/hostpath/README.md
```diff
@@ -7,52 +7,70 @@
...
+$ #endpoint=tcp://127.0.0.1:10000 # tcp
+```
+
 ### Start Hostpath driver
 ```
-$ sudo ./bin/hostpathplugin --endpoint tcp://127.0.0.1:10000 --nodeid CSINode -v=5
+$ sudo ./bin/hostpathplugin --endpoint $endpoint --nodeid CSINode -v=5
 ```
 
 ### Test using csc
```
xing-yang, line 21: https://github.com/rexray/gocsi is no longer supported. We should remove this entire section that relies on it.

astraw99: Currently the `csc` tool works ok.
If we remove this section, is there any other tool to use?

astraw99:
/remove-lifecycle stale
@xing-yang PTAL thanks.

pkg/hostpath/README.md
nixpanic: `csi-sanity` can be used, it is part of the [csi-test](https://github.com/kubernetes-csi/csi-test) repository.

astraw99: Got it, will try the tool and update the doc.
Thanks.

PR #480, opened 2023-12-05
fix missing published target paths when republish the ephemeral volume

carlory:
**What type of PR is this?**

/kind bug

calling `createVolume` twice will reset volume state. It will cause `NodeUnpublishVolume` to be unable to perform unmount operation.

FYI:

- https://github.com/carlory/csi-driver-host-path/blob/4ffa1fc0586fa41cad7ce5861c1e5a98227d8bad/pkg/hostpath/hostpath.go#L223
- https://github.com/carlory/csi-driver-host-path/blob/4ffa1fc0586fa41cad7ce5861c1e5a98227d8bad/pkg/hostpath/nodeserver.go#L230

Ref https://github.com/kubernetes/kubernetes/issues/120080

```release-note
Fix missing published target paths when republish the ephemeral volume
```

### DIFF
--- a/pkg/hostpath/nodeserver.go
+++ b/pkg/hostpath/nodeserver.go
@@ -69,15 +69,18 @@ func (hp *hostPath) NodePublishVolume(ctx context.Context, req *csi.NodePublishV
 	if ephemeralVolume {
 		volID := req.GetVolumeId()
 		volName := fmt.Sprintf("ephemeral-%s", volID)
-		kind := req.GetVolumeContext()[storageKind]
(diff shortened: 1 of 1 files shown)


carlory:
@xing-yang 

the solution to [this issue](https://github.com/kubernetes/kubernetes/issues/120080) depends on the release version which contains the pr. I want to confirm if we need check-pick it into the previous release and cut a patched release, or wait for the next release which contains VAC feature implementation. I'd pefer to the frist option. what do you think?

xing-yang:
Go ahead and cherry-pick it.

PR #481, opened 2023-12-11
hostpath driver implements the ControllerModityVolume feature

### LINKED ISSUE #479: KEP-3751 - VAC Hostpath Driver Implementation
Currently in https://github.com/kubernetes-csi/csi-driver-host-path/blob/master/pkg/hostpath/controllerserver.go#L496 ControllerModifyVolume is marked as not implemented.

This implementation will help us test the alpha feature.

carlory:
**What type of PR is this?**

/kind feature

Fixes #479

```release-note
hostpath driver implements the ControllerModityVolume feature
```

### DIFF
--- a/pkg/hostpath/controllerserver.go
+++ b/pkg/hostpath/controllerserver.go
@@ -33,6 +33,7 @@ import (
 	"google.golang.org/grpc/status"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog/v2"
 
 	"github.com/kubernetes-csi/csi-driver-host-path/pkg/state"
@@ -46,6 +47,17 @@ func (hp *hostPath) CreateVolume(ctx context.Context, req *csi.CreateVolumeReque
 	if err := hp.validateControllerServiceRequest(csi.ControllerServiceCapability_RPC_CREATE_DELETE_VOLUME); err != nil {
 		glog.V(3).Infof("invalid create volume req: %v", req)
 		return nil, err
+	}
+
+	if len(req.GetMutableParameters()) > 0 {
+		if err := hp.validateControllerServiceRequest(csi.ControllerServiceCapability_RPC_MODIFY_VOLUME); err != nil {
+			glog.V(3).Infof("invalid create volume req: %v", req)
+			return nil, err
+		}
+		// Check if the mutable parameters are in the accepted list
+		if err := hp.validateVolumeMutableParameters(req.MutableParameters); err != nil {
+			return nil, err
+		}
 	}
 
 	// Check arguments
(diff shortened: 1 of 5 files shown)


pkg/hostpath/controllerserver.go
```diff
@@ -48,6 +49,26 @@ func (hp *hostPath) CreateVolume(ctx context.Context, req *csi.CreateVolumeReque
...
+			accepts := sets.New(hp.config.AcceptedMutableParameterNames...)
+			unsupported := []string{}
+			for k := range req.MutableParameters {
+				if !accepts.Has(k) {
+					unsupported = append(unsupported, k)
+				}
+			}
+			if len(unsupported) > 0 {
+				return nil, status.Errorf(codes.InvalidArgument, "Parameters %v are not supported", unsupported)
+			}
```
gnufied, line 68: nit: may be validation of these parameters can be moved to a common function or something.

carlory: @gnufied updated.

pkg/hostpath/controllerserver.go
```diff
@@ -493,8 +505,36 @@ func (hp *hostPath) ControllerGetVolume(ctx context.Context, req *csi.Controller
...
+	// driver might use more fine-grained locking.
+	hp.mutex.Lock()
+	defer hp.mutex.Unlock()
+
+	// Get the volume
+	_, err := hp.state.GetVolumeByID(req.VolumeId)
+	if err != nil {
+		return nil, status.Error(codes.NotFound, err.Error())
+	}
+
```
carlory, line 536: /cc @sunnylovestiramisu @msau42 

I'm not sure whether mutable parameters should be store in the [volume state] https://github.com/kubernetes-csi/csi-driver-host-path/blob/master/pkg/state/state.go#L39)? It seems useless to do that.

sunnylovestiramisu: We can add a test parameter here but make it really clear that it is for testing(ModifyVolumeTestingField etc.), also comment out the field so that others do not use it.

pkg/hostpath/controllerserver.go
```diff
@@ -493,8 +505,36 @@ func (hp *hostPath) ControllerGetVolume(ctx context.Context, req *csi.Controller
 	}, nil
 }
 
-func (hp *hostPath) ControllerModifyVolume(ctx context.Context, request *csi.ControllerModifyVolumeRequest) (*csi.ControllerModifyVolumeResponse, error) {
-	return nil, fmt.Errorf("unimplemented")
+func (hp *hostPath) ControllerModifyVolume(ctx context.Context, req *csi.ControllerModifyVolumeRequest) (*csi.ControllerModifyVolumeResponse, error) {
```
sunnylovestiramisu, line 508: Should we add some unit tests coverage?

carlory: @sunnylovestiramisu @gnufied unit tests are added in https://github.com/kubernetes-csi/csi-driver-host-path/compare/28e3543ae06533f5edee2fe8d2e0c0fb76f4b021..218a334eb336d648dcfb4439037ce65f93c1582d

pkg/hostpath/controllerserver.go
```diff
@@ -782,6 +822,25 @@ func convertSnapshot(snap state.Snapshot) *csi.ListSnapshotsResponse {
...
+
+	accepts := sets.New(hp.config.AcceptedMutableParameterNames...)
+	unsupported := []string{}
+	for k := range params {
+		if !accepts.Has(k) {
+			unsupported = append(unsupported, k)
+		}
+	}
+	if len(unsupported) > 0 {
+		return status.Errorf(codes.InvalidArgument, "Parameters %v are not unsupported", unsupported)
```
sunnylovestiramisu, line 839: Do you mean "is not supported"?

carlory: If a parmater is not in the accepted list, it's will be reject because the parameter is supported to be modified. the accept list is used to simulate the request reject by csi-driver. the "are" word means that no more than one paramters.

how about this?

```go
 status.Errorf(codes.InvalidArgument, "invalid parameters: %v", unsupported)
```

sunnylovestiramisu: This looks better

sunnylovestiramisu:
/lgtm 

@gnufied Do you have an idea of the mock client you used for testing before?

PR #501, opened 2024-02-09
Update deployment structure and sidecar versions for Kubernetes 1.29

RaunakShah:
**What type of PR is this?**

/kind cleanup

This PR updates the directory structure of ./deploy. Also update the sidecars to their latest releases, corresponding to kubernetes 1.29.

Fixes #

```release-note
Update deploy directory structure and update sidecar releases, corresponding to Kubernetes 1.29.
```


xing-yang:
I think we still need to keep 1.27 as we need to test 3 releases (1.27, 1.28, and latest). See the following folder structure for 1.28:

https://github.com/kubernetes-csi/csi-driver-host-path/tree/release-1.12/deploy

RaunakShah:
@xing-yang I've moved the `1.26` files to `1.27`. The git diff shows that `1.27` files were deleted and then `1.26` were renamed to `1.27`. 
I think it shows this way because `1.27` and `latest` were only symlinks to `1.26`.

xing-yang:
> @xing-yang I've moved the 1.26 files to 1.27. The git diff shows that 1.27 files were deleted and then 1.26 were renamed to 1.27.
I think it shows this way because 1.27 and latest were only symlinks to 1.26.

You are right.  Thanks!

PR #526, opened 2024-05-24
update release-tools 

humblec:
> /kind cleanup

-->
```release-note
NONE
```

### DIFF
--- a/release-tools/.github/dependabot.yaml
+++ b/release-tools/.github/dependabot.yaml
@@ -1,0 +1,12 @@
+version: 2
+enable-beta-ecosystems: true
+updates:
+- package-ecosystem: "github-actions"
+  directory: "/"
+  schedule:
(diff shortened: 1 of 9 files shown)


pohly:
Are release-tools updates now done manually in each repo on an as-needed basis?

I had a script which did it for several repos at once and also produced better PR descriptions.

humblec:
> Are release-tools updates now done manually in each repo on an as-needed basis?
> 

Came to this after fixing the code-spell error like  in #255.

> I had a script which did it for several repos at once and also produced better PR descriptions.

Actually speaking this PR is half baked with that same script.  That is commits were prepared by the same, however PR was not raised by it. I can try to use the same for entire process too.

pohly:
I've not used that script in a while and we never had an "official" version of it. I was just wondering.

/approve
/lgtm

PR #534, opened 2024-06-18
switch from glog to klog

huww98:
**What type of PR is this?**
>
> /kind api-change
> /kind bug

/kind cleanup

> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Replace glog with klog in all go source files.
Fix several compile errors. (`if v5` => `if v5.Enabled()`)
Add "klog.InitFlags(nil)" and remove flag.Set("logtostderr", "true"), which is now the default.
Run go mod tidy + go mod vendor.

Fixes https://github.com/kubernetes/kubernetes/issues/125539

/cc @jsafrane 

```release-note
Switched logging library from glog to klog
```


huww98:
Should fix https://github.com/kubernetes/kubernetes/issues/125539

carlory:
@xing-yang before cutting a release, please merge it. I will update the host path manifest in https://github.com/kubernetes/kubernetes/pull/124842.

@jsafrane BTW, should we update the hostpath image tag to v1.14.0 before cutting a release? 

- https://github.com/kubernetes-csi/csi-driver-host-path/blob/master/deploy/kubernetes-1.27-test/hostpath/csi-hostpath-plugin.yaml#L34

jsafrane:
> @jsafrane BTW, should we update the hostpath image tag to v1.14.0 before cutting a release?

I don't know. In the history I can see we updated the number *after* a release was tagged. Did we do it wrong?

carlory:
IMO, If we update the deployment file before releasing the version, then we don't need to do any extra work after releasing the version.

cc @xing-yang what do you think?

PR #543, opened 2024-07-15
Allow ignoring unreadable files in `tar`-based snapshots

### LINKED ISSUE #542: Snapshot fails if a file is deleted concurrently
In CNPG E2E tests, we take several snapshots of a running PostgreSQL cluster. Sometimes, the snapshot fails because a file is deleted between when tar gathers the file list and when it is read. This is an example of the failure:

```
can't execute snapshot backup: Failed to create snapshot: failed to take snapshot of the volume a62e8015-3ab3-11ef-813a-829810a7d821: "rpc error: code = Unknown desc = failed create snapshot: exit status 1: tar: ./pgdata/base/1/pg_internal.init: No such file or directory\ntar: error exit delayed from previous errors\n"
```

The file `pg_internal.init` is the PostgreSQL relation cache and is deleted and recreated during PostgreSQL startup.

leonardoce:
Snapshots are implemented using `tar` for file system volumes. With this patch, users can set the `ignoreFailedRead` parameter in the VolumeSnapshotClass parameter to `true`, which results in the `--ignore-failed-read` option to be passed to `tar`.

This enhancement is particularly useful for end-to-end testing of applications that may delete files initially included when the tar process started and retrieved the file list. By ignoring unreadable files, the snapshot process can finish successfully.

**What type of PR is this?**
>
> /kind api-change
/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #542 

```release-note
The `ignoreFailedRead` parameter in the VolumeSnapshotClass, when set to `true`, results in the `--ignore-failed-read` option to be passed to `tar`.
```

### DIFF
--- a/pkg/hostpath/hostpath.go
+++ b/pkg/hostpath/hostpath.go
@@ -380,17 +380,25 @@ func (hp *hostPath) getAttachCount() int64 {
 	return count
 }
 
-func (hp *hostPath) createSnapshotFromVolume(vol state.Volume, file string) error {
-	var cmd []string
+func (hp *hostPath) createSnapshotFromVolume(vol state.Volume, file string, opts ...string) error {
+	var args []string
+	var cmdName string
 	if vol.VolAccessType == state.BlockAccess {
 		klog.V(4).Infof("Creating snapshot of Raw Block Mode Volume")
-		cmd = []string{"cp", vol.VolPath, file}
+		cmdName = "cp"
+		args = []string{vol.VolPath, file}
 	} else {
-		klog.V(4).Infof("Creating snapshot of Filsystem Mode Volume")
-		cmd = []string{"tar", "czf", file, "-C", vol.VolPath, "."}
+		klog.V(4).Infof("Creating snapshot of Filesystem Mode Volume")
+		cmdName = "tar"
+		opts = append(
+			[]string{"czf", file, "-C", vol.VolPath},
+			opts...,
+		)
+		args = []string{"."}
 	}
 	executor := utilexec.New()
-	out, err := executor.Command(cmd[0], cmd[1:]...).CombinedOutput()
+	optsAndArgs := append(opts, args...)
+	out, err := executor.Command(cmdName, optsAndArgs...).CombinedOutput()
 	if err != nil {
 		return fmt.Errorf("failed create snapshot: %w: %s", err, out)
 	}
--- a/pkg/hostpath/options.go
+++ b/pkg/hostpath/options.go
@@ -1,0 +1,52 @@
+/*
+Copyright 2024 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package hostpath
+
+import (
+	"fmt"
+	"strconv"
+
+	"github.com/kubernetes-csi/csi-driver-host-path/pkg/state"
+)
+
+// ignoreFailedReadParameterName is a parameter that, when set to true,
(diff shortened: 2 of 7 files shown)


carlory:
/lgtm
/assign @xing-yang @jsafrane 
Please take another look.

pkg/hostpath/options.go
```diff
@@ -0,0 +1,56 @@
...
+	"fmt"
+
+	"github.com/kubernetes-csi/csi-driver-host-path/pkg/state"
+)
+
+// snapshotOptionsMountedParameterName is the name of the parameters containing
+// the set of additional options to be passed to `tar` when creating a snapshot
+// for a mounted volume. The options should be passed as a JSON formatted string
+// array
+const snapshotOptionsMountedParameterName = "hostpath.csi.k8s.io/snapshotOptionsMounted"
```
jsafrane, line 30: nit: `hostpath.csi.k8s.io/snapshotOptionsFilesystem`.
In Kubernetes world (= VolumeSnapshotClass), the volume mode is either `Filesystem` or `Block`, `Mount` is a CSI term.

pkg/hostpath/options.go
```diff
@@ -0,0 +1,56 @@
...
+		optsString = parameters[snapshotOptionsMountedParameterName]
+	}
+
+	if len(optsString) == 0 {
+		return nil, nil
+	}
+
+	var opts []string
+	if err := json.Unmarshal([]byte(optsString), &opts); err != nil {
+		return nil, fmt.Errorf("JSON parsing error in parameter: %w", err)
```
jsafrane, line 52: Please print the parameter name

pkg/hostpath/options_test.go
```diff
@@ -0,0 +1,154 @@
+/*
+Copyright 2021 The Kubernetes Authors.
```
jsafrane, line 2: 2024 \o/

pkg/hostpath/options.go
```diff
@@ -0,0 +1,56 @@
+/*
+Copyright 2021 The Kubernetes Authors.
```
jsafrane, line 2: 2024

pkg/hostpath/hostpath.go
```diff
@@ -380,17 +380,25 @@ func (hp *hostPath) getAttachCount() int64 {
...
+		cmdName = "cp"
+		args = []string{vol.VolPath, file}
 	} else {
-		klog.V(4).Infof("Creating snapshot of Filsystem Mode Volume")
-		cmd = []string{"tar", "czf", file, "-C", vol.VolPath, "."}
+		klog.V(4).Infof("Creating snapshot of Filesystem Mode Volume")
+		cmdName = "tar"
+		opts = append(
+			[]string{"czf", file, "-C", vol.VolPath},
+			opts...,
```
jsafrane, line 395: I hope that passing of unsanitized parameters from user to `tar` is safe enough. We do it with `mount` already.

leonardoce: Yeah, that's a really good point. What should we do about it?
We could change our approach and allow injecting only some known options...

jsafrane: GNU tar supports:
> `--to-command=COMMAND`
              Pipe extracted files to COMMAND.  The argument is the pathname of an external program, optionally with command line arguments.  The program will be invoked and the contents of the file being extracted supplied to it on its standard input.  Additional data will be supplied via the following environment variables [...]

To me it seems it's very unsafe to pass an unsanitized input. Even when csi-driver-hostpath is mostly used for testing, we should not ship code that can be unsafe when used by someone who mis-uses it it a real cluster.

jsafrane: /hold

leonardoce: @jsafrane, what do you think of adding a parameter just for `--ignore-failed-read`?
Something like `hostpath.csi.k8s.io/ignoreFailedRead: true`.
That would be used only for snapshots of filesystem volumes.

jsafrane: You don't need to prefix VolumeSnapshotClass parameters, simple `ignoreFailedRead: "true"` is OK.

leonardoce: Great. I'm reworking the patch to handle that.

leonardoce: I'm done. Now it's definitely better, thank you!

PR #577, opened 2025-01-01
Tighter checks on CreateVolume CapacityRange

ebblake:
The CSI spec is clear that when CapacityRange is supplied, neither RequiredBytes nor LimitBytes may be negative, and at least one of the two values must be non-zero.  It is also impossible to create a volume that requires more bytes than a non-zero limit.  However, before this patch, we were not taking the limit into account.

Note that since we do not check whether hp.createVolume() rounds capacity up during the creation, we cannot detect the case where that rounding would violate limit.

See also https://github.com/kubernetes-csi/csi-test/pull/568 which will test these changes.

**What type of PR is this?**
/kind bug

Add better validation of the CapacityRange parameter to CreateVolume, to set a better example to other CSI drivers.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/hostpath/controllerserver.go
+++ b/pkg/hostpath/controllerserver.go
@@ -97,12 +97,23 @@ func (hp *hostPath) CreateVolume(ctx context.Context, req *csi.CreateVolumeReque
 		requestedAccessType = state.MountAccess
 	}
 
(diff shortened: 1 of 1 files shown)


ebblake:
Ping

xing-yang:
Never mind.  I added a release note.

PR #569, opened 2024-10-15
Implement CSI SnapshotMetadata service

PrasadG193:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

This PR implements SnapshotMetadata service from the [CSI specs](https://github.com/container-storage-interface/spec/blob/master/csi.proto#L130). This service in combination with [external-snapshot-metadata](https://github.com/kubernetes-csi/external-snapshot-metadata) will be used to compare and query changed block metadata between two snapshots.
Right now, the service will be disabled by default till `external-snapshot-metadata` alpha is release.

The changes in deploy script to add `external-snapshot-metadata` sidecar will be added after alpha release.

#### NOTE
This is a sample implementation of SnapshotMetadata intended to be used for demo and CI testing purpose only.
This should not be used in production or as an example about how to write a real driver.

##### Test plan
Complete test plan is documented at - https://gist.github.com/PrasadG193/1fab082f03bbdb7db25a41ed17b7e7c1

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
Add sample implementation of CSI SnapshotMetadata service 
```

PR #579, opened 2025-01-08
fix: consider only non-zero blocks for GetMetadataAllocated

iPraveenParihar:
**What type of PR is this?**
/kind bug

The `GetMetadataAllocated` method previously returned all blocks,
including zero-filled ones (usually, unallocated/unused).
This commit updates the method to return only allocated (non-zero) blocks.

Fixes #

```release-note
Fixed: Considered only non-zero blocks for GetMetadataAllocated.
```

### DIFF
--- a/pkg/hostpath/snapshotmetadata.go
+++ b/pkg/hostpath/snapshotmetadata.go
@@ -118,6 +118,7 @@ func (cb *fileBlockReader) getChangedBlockMetadata(ctx context.Context) ([]*csi.
 	blockIndex := cb.offset / cb.blockSize
 	sBuffer := make([]byte, cb.blockSize)
 	tBuffer := make([]byte, cb.blockSize)
+	zeroBlock := make([]byte, cb.blockSize)
 	eofBaseFile, eofTargetFile := false, false
 
 	changedBlocks := []*csi.BlockMetadata{}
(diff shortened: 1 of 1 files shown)


pkg/hostpath/snapshotmetadata.go
```diff
@@ -143,15 +144,19 @@ func (cb *fileBlockReader) getChangedBlockMetadata(ctx context.Context) ([]*csi.
...
 			}
-			// if VARIABLE_LENGTH type is enabled, return blocks extend instead of individual blocks.
-			blockMetadata := createBlockMetadata(blockIndex, cb.blockSize)
-			if extendBlock(changedBlocks, csi.BlockMetadataType(cb.blockMetadataType), blockIndex, cb.blockSize) {
-				changedBlocks[len(changedBlocks)-1].SizeBytes += cb.blockSize
-				cb.offset += cb.blockSize
-				blockIndex++
-				continue
+			// return only allocated blocks.
+			if blockChanged(zeroBlock, tBuffer[:targetReadBytes]) {
```
nixpanic, line 148: Does this work when a change wrote a whole lot of zero bytes over existing data to the volume?

nixpanic: I guess it should be fine, as `cb.base == nil` here.

iPraveenParihar: This is for `GetMetadataAllocated`, where the backup application requires only the allocated data from the snapshot. During restoration, any uninitialized bytes will be filled with zeroes anyway.

Rakshith-R (approved):
Looks good to me

PrasadG193 (approved):
/lgtm

xing-yang:
Please add a release note.

Rakshith-R:
> we can remove hold after a release note is added.

Hey @xing-yang 

The API implementation pr went was merged in december 2024 
https://github.com/kubernetes-csi/csi-driver-host-path/pull/569#issuecomment-2551950755

The last release was in september 2024 https://github.com/kubernetes-csi/csi-driver-host-path/releases/tag/v1.15.0 

This pr would fix the implementation before it is present in a release. 
Do we still need a release note in this case where the bug is never present in a previous release?

iPraveenParihar:
added release note. PTAL @xing-yang 😃

PR #609, opened 2025-06-09
{WIP} Add support for running hostpath volumes in rwx volume mode

gnufied:
This is just a way to run hostpath driver in rwx mode. 

I am currently using a NFS share to share the state.json file between multiple nodes. Probably not most elegant.


gnufied:
I am going to close this. I think a better way of doing this is to use kind and RWX volumes. closing for now.

Issue #651, opened 2026-04-10
Enhance distributed deployment with snapshot and resize support for multi-node testing


mnencia:
## Problem

The standard StatefulSet deployment (replicas=1) binds ALL volumes to a single node, making multi-node test scenarios impossible. This has been a recurring request:
- #211 ("Use hostpath.csi.k8s.io across multiple node pools")
- #49 ("Hostpath across multiple nodes")
- #392 (distributed snapshotting attempt, went stale and was closed by the triage bot)

The existing `deploy/kubernetes-distributed/` DaemonSet deployment partially addresses this: it runs the driver on every node with `csi-provisioner --node-deployment=true`, enabling per-node volume provisioning with topology and capacity tracking. However, it only includes the provisioner sidecar, so **snapshots and volume expansion are not supported** in the distributed deployment.

## Proposal

Enhance the distributed deployment to include the **csi-snapshotter** and **csi-resizer** sidecars, using their respective `--node-deployment` modes so each instance only handles operations for volumes local to its node.

### Snapshotter: upstream support exists

The external-snapshotter already supports distributed snapshotting (kubernetes-csi/external-snapshotter#585, merged Dec 2021). It requires two coordinating flags on two separate components:

1. The **common snapshot-controller** (a central Deployment) must be deployed with `--enable-distributed-snapshotting=true`. This makes it check PV node affinity and label each VolumeSnapshotContent with `snapshot.storage.kubernetes.io/managed-by=<nodeName>`.
2. The **csi-snapshotter sidecar** (per-node, in the DaemonSet) must be deployed with `--node-deployment=true`. This makes it only reconcile VolumeSnapshotContent objects whose label matches its node.

Without the snapshot-controller flag, the labeling never activates and the per-node sidecars would see nothing to process.

The integration still needs to be built and tested against current sidecar versions (the standard deployment uses `csi-snapshotter:v8.4.0`). The distributed `deploy.sh` would need to deploy (or patch) the common snapshot-controller with the non-default flag, add a VolumeSnapshotClass, and handle RBAC bindings. Notably, the DaemonSet currently uses `serviceAccountName: csi-provisioner`, so upstream snapshotter RBAC (which binds to a `csi-snapshotter` service account) would need to be adapted to bind to `csi-provisioner` instead.

### Resizer: upstream work needed

[...]

PR #392, opened 2023-02-10
add distributed snapshotter

denisok:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:

This adds snapshot capabilities to the distributed deployment.

Don't know if this fill be useful, I did before: https://github.com/kubernetes/minikube/pull/15829 . To test possibility to have snapshots for multi-node.

**Which issue(s) this PR fixes**:

It is related to #211  . Adds distributed snapshotting

Fixes #328

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
Enable distributed snapshotting for multi-node deployments
```

PR #653, opened 2026-04-12
Add distributed snapshotting support to kubernetes-distributed deployment

mnencia:
**What type of PR is this?**

/kind feature

The distributed DaemonSet deployment (`deploy/kubernetes-distributed/`) only includes the csi-provisioner sidecar. This adds the csi-snapshotter with `--node-deployment=true`, so snapshots work on each node's local volumes.

To support multiple sidecar RBAC roles, this introduces a unified ServiceAccount (`csi-hostpathplugin-sa`) with explicit ClusterRoleBindings for both the provisioner and snapshotter roles, following the same pattern used by the `kubernetes-latest` deployment.

The deploy script now handles the full snapshot infrastructure: CRD installation, snapshot-controller deployment with `--enable-distributed-snapshotting=true`, and node-reader RBAC (required for distributed snapshotting but commented out in the upstream snapshot-controller RBAC). If the snapshot-controller was already deployed (e.g., by prow.sh) without the flag, the script patches it. The destroy script cleans up all of these resources.

Snapshot E2E tests are enabled via `snapshotDataSource` and `SnapshotClass` in test-driver.yaml.

Part of #651

Builds on prior work in #392 by @denisok, which went stale before merging.

The distributed snapshotting feature in external-snapshotter (kubernetes-csi/external-snapshotter#585) requires coordination between two components:
1. The common snapshot-controller must run with `--enable-distributed-snapshotting=true` to label VolumeSnapshotContent objects with node affinity
2. The per-node csi-snapshotter sidecar must run with `--node-deployment=true` to filter by those labels

The upstream snapshot-controller RBAC has Node read permissions commented out. The deploy script applies them via a separate `snapshot-controller-node-reader` ClusterRole.

Tested on a 3-node Kind cluster: provisioning on different workers, snapshot creation and deletion all work correctly.

```release-note
The kubernetes-distributed deployment now supports volume snapshots on per-node volumes.
```

### DIFF
--- a/deploy/kubernetes-distributed/deploy.sh
+++ b/deploy/kubernetes-distributed/deploy.sh
@@ -122,6 +122,10 @@ function version_gt() {
 CSI_PROVISIONER_RBAC_YAML="https://raw.githubusercontent.com/kubernetes-csi/external-provisioner/$(rbac_version "${BASE_DIR}/hostpath/csi-hostpath-plugin.yaml" csi-provisioner false)/deploy/kubernetes/rbac.yaml"
 : ${CSI_PROVISIONER_RBAC:=https://raw.githubusercontent.com/kubernetes-csi/external-provisioner/$(rbac_version "${BASE_DIR}/hostpath/csi-hostpath-plugin.yaml" csi-provisioner "${UPDATE_RBAC_RULES}")/deploy/kubernetes/rbac.yaml}
 
+SNAPSHOTTER_RBAC_RELATIVE_PATH="csi-snapshotter/rbac-csi-snapshotter.yaml"
+CSI_SNAPSHOTTER_RBAC_YAML="https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/$(rbac_version "${BASE_DIR}/hostpath/csi-hostpath-plugin.yaml" csi-snapshotter false)/deploy/kubernetes/${SNAPSHOTTER_RBAC_RELATIVE_PATH}"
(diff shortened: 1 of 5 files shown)


deploy/kubernetes-distributed/deploy.sh
```diff
@@ -169,6 +173,71 @@ EOF
     run kubectl apply --kustomize "${TEMP_DIR}"
 done
 
+# Install snapshot CRDs and snapshot-controller if not already present.
```
jsafrane: I think ‎`kubernetes-distributed` should be in sync with `deploy.sh`. Either both of them should install snapshot CRDs and the controller or none of them. I personally prefer keep them as they are, without the CRD and controller installation.

mnencia: Done. Stripped the snapshot CRDs and snapshot-controller install from `deploy.sh` and the matching cleanup from `destroy.sh`. The README now documents that `snapshot-controller --enable-distributed-snapshotting=true` (and the `VolumeSnapshot*` CRDs) is a prerequisite the operator installs separately, the same way `deploy/util/deploy-hostpath.sh` works.

While I was at it I also reverted the `snapshotDataSource: true` flip in `deploy/kubernetes-distributed/test-driver.yaml`. That was the cause of the `pull-kubernetes-csi-csi-driver-host-distributed-on-kubernetes-1-28` failure: with the snapshotter sidecar present but no fix for cross-node snapshot restore, the snapshot E2E suite deadlocks at `PreBind` waiting for `bindTimeout`. Cross-node restore is handled by a follow-up controller (`csi-topology-coordinator`) that I will open as a separate PR; the test-driver flag flips back on there once the controller is in place.

PR #658, opened 2026-04-24
Fix inverted error check in DeleteSnapshot

carterpewpew:
**What type of PR is this?**

/kind bug

`GetSnapshotByID` returns `(Snapshot{}, NotFound)` when a snapshot is not found, so the original condition `err != nil && snapshot.GroupSnapshotID != ""` could never evaluate to true the zero-value `Snapshot` always has `GroupSnapshotID == ""`. The guard needs `err == nil` to inspect the fields of a successfully retrieved snapshot.

As a result, individual snapshots belonging to a `VolumeGroupSnapshot` could be deleted via `DeleteSnapshot`, breaking group snapshot consistency guarantees.

This PR also adds test coverage for `DeleteSnapshot`.

```release-note
NONE
```

### DIFF
--- a/pkg/hostpath/controllerserver.go
+++ b/pkg/hostpath/controllerserver.go
@@ -649,7 +649,12 @@ func (hp *hostPath) DeleteSnapshot(ctx context.Context, req *csi.DeleteSnapshotR
 	defer hp.mutex.Unlock()
 
 	// If the snapshot has a GroupSnapshotID, deletion is not allowed and should return InvalidArgument.
-	if snapshot, err := hp.state.GetSnapshotByID(snapshotID); err != nil && snapshot.GroupSnapshotID != "" {
+	snapshot, err := hp.state.GetSnapshotByID(snapshotID)
+	if err != nil {
+		if status.Code(err) != codes.NotFound {
+			return nil, err
+		}
+	} else if snapshot.GroupSnapshotID != "" {
 		return nil, status.Errorf(codes.InvalidArgument, "Snapshot with ID %s is part of groupsnapshot %s", snapshotID, snapshot.GroupSnapshotID)
 	}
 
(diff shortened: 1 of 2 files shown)


carlory:
/ok-to-test
/cc @xing-yang @nixpanic 
related to https://github.com/kubernetes-csi/csi-driver-host-path/pull/399

pkg/hostpath/controllerserver.go
```diff
@@ -649,7 +649,7 @@ func (hp *hostPath) DeleteSnapshot(ctx context.Context, req *csi.DeleteSnapshotR
 	defer hp.mutex.Unlock()
 
 	// If the snapshot has a GroupSnapshotID, deletion is not allowed and should return InvalidArgument.
-	if snapshot, err := hp.state.GetSnapshotByID(snapshotID); err != nil && snapshot.GroupSnapshotID != "" {
+	if snapshot, err := hp.state.GetSnapshotByID(snapshotID); err == nil && snapshot.GroupSnapshotID != "" {
 		return nil, status.Errorf(codes.InvalidArgument, "Snapshot with ID %s is part of groupsnapshot %s", snapshotID, snapshot.GroupSnapshotID)
 	}
```
nixpanic, line 659: Thanks for fixing this!

There should be an `err != nil` check as well. We shouldn't try deleting the snapshot if `hp.state.GetSnapshotByID()` returned an error.

carterpewpew: Addressed, PTAL. Non-NotFound errors from GetSnapshotByID are now returned instead of silently ignored. NotFound still falls through to the idempotent cleanup path per the CSI spec. Thanks for the review !

nixpanic: Looks good, thanks!

nixpanic (commented):
/lgtm

PR #581, opened 2025-02-04
Update deployment structure and sidecar versions for Kubernetes 1.32

leonardoce:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespaces from that line:
>
> /kind api-change
> /kind bug

/kind cleanup

> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

**What this PR does / why we need it**:

This PR updates the directory structure of ./deploy. Also update the sidecars to their latest releases, corresponding to Kubernetes 1.32.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
Update deploy directory structure and update sidecar releases, corresponding to Kubernetes 1.32.
```

PR #665, opened 2026-06-25
deploy: add Kubernetes 1.36 manifests

darshansreenivas:
  Adds deployment manifests for Kubernetes 1.36, including both standard (`kubernetes-1.36`) and test (`kubernetes-1.36-test`) configurations with all required
  CSI sidecar yamls (attacher, provisioner, resizer, snapshotter, plugin, etc.).

  NONE

  Adds deployment manifests for Kubernetes 1.36, including both standard (`kubernetes-1.36`) and test (`kubernetes-1.36-test`) configurations with all required
  CSI sidecar yamls (attacher, provisioner, resizer, snapshotter, plugin, etc.).

  NONE


xing-yang:
Please use this PR https://github.com/kubernetes-csi/csi-driver-host-path/pull/581 as an example. That PR adds deployment structure for 1.32 where 1.32 is the latest. We support 3 versions and the other two versions are 1.30 and 1.31 there.

In this case, you are adding deployment structure for 1.36 which should become the latest. We support 3 versions. So the other two versions should be 1.34 and 1.35.

PR #678, opened 2026-07-27
Implement volume health changes in CSI driver

gnufied:
Implement a mechanism for manually triggering health errors

We also need to remove health-monitor e2e for now and they will be restored once external-health-monitor release has been tagged.

```release-note
Add support for newer Volume Health RPCs
```


xing-yang:
In the PR description, can you add that health-monitor tests have been removed but will be added back after this PR is merged?

cmd/healthctl/main.go
```diff
@@ -0,0 +1,332 @@
...
+		os.Exit(2)
+	}
+	return scope
+}
+
+func runMarkVolumeUnhealthy(args []string) int {
+	fs := flag.NewFlagSet("mark-volume-unhealthy", flag.ExitOnError)
+	scope := fs.String("scope", "both", "which side reports unhealthy: both, controller, or node")
+	status := fs.String("status", "DEGRADED", "volume health status (one of: DEGRADED, INACCESSIBLE, DATA_LOSS)")
+	reason := fs.String("reason", "manually-marked", "short CamelCase reason for the condition")
```
xing-yang: "manually-marked" is not CamelCase.

cmd/healthctl/main.go
```diff
@@ -0,0 +1,332 @@
...
+		return 1
+	}
+	fmt.Printf("volume %s %s-scope marker cleared\n", volID, sc)
+	return 0
+}
+
+func runMarkStorageUnhealthy(args []string) int {
+	fs := flag.NewFlagSet("mark-storage-unhealthy", flag.ExitOnError)
+	status := fs.String("status", "STORAGE_DEGRADED", "storage health status (one of: STORAGE_UNREACHABLE, STORAGE_DEGRADED)")
+	reason := fs.String("reason", "manually-marked", "short CamelCase reason for the condition")
```
xing-yang: Here too.  "manually-marked" is not CamelCase.  Please check other places as well.

pkg/hostpath/controllerserver.go
```diff
@@ -894,6 +872,87 @@ func (hp *hostPath) validateVolumeMutableParameters(params map[string]string) er
...
+	startIdx := int64(1)
+	if req.GetStartingToken() != "" {
+		s, err := strconv.ParseInt(req.GetStartingToken(), 10, 64)
+		if err != nil || s < 1 {
+			return nil, status.Error(codes.Aborted, "The type of startingToken should be a positive integer")
+		}
+		startIdx = s
+	}
+
+	total := int64(len(volIDs))
```
torredil, line 927: there is an edge case here where if `startingToken` is larger than the number of entries (`total`), then make function below will panic with a negative slice capacity, worth returning aborted error similar to how the invalid token case is handled above:

```suggestion
	total := int64(len(volIDs))
	if startIdx-1 > total {
		return nil, status.Error(codes.Aborted, "startingToken is beyond the end of the list")
	}
```

torredil (commented):
One suggestion for an edge case in pagination logic, otherwise this looks pretty good to me. Removing the external health monitor deployment for now makes sense

deploy/util/deploy-hostpath.sh
```diff
@@ -190,7 +187,7 @@ run () {
 
 # rbac rules
 echo "applying RBAC rules"
-components=(CSI_PROVISIONER CSI_ATTACHER CSI_SNAPSHOTTER CSI_RESIZER CSI_EXTERNALHEALTH_MONITOR)
+components=(CSI_PROVISIONER CSI_ATTACHER CSI_SNAPSHOTTER CSI_RESIZER)
 if snapshot_metadata; then
     components+=(CSI_SNAPSHOT_METADATA)
```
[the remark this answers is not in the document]
gnufied: Added wording in description about why we are removing this.

deploy/kubernetes-1.34/hostpath/csi-hostpath-plugin.yaml
```diff
@@ -31,24 +31,6 @@ subjects:
...
-  name: external-health-monitor-controller-runner
-subjects:
-- kind: ServiceAccount
-  name: csi-hostpathplugin-sa
-  namespace: default
----
-apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRoleBinding
 metadata:
   labels:
```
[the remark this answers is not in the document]
gnufied: Added wording in description about why we are removing this.

cmd/healthctl/main.go
gnufied: Fixed.

